PR #78, opened 2021-02-19
Fix/Improve keyboard navigation for login dialog

### LINKED ISSUE #70: Login doesn't allow user to tab between inputs
**Describe the bug**
Login doesn't allow tabbing between fields

**To Reproduce**
1. Go to login screen
2. Select username
3. Active element doesn't change

**Expected behavior**
Tab switches user selection between username, password and login fields

**General information:**
 - Distribution: Arch
 - Installation method [e. g. built from source, installed from Flathub...]: Flathub
 - Version [e.g. 0.1.0]: 0.1.7

Douile:
This PR fixes tab navigation not working on the login dialog. It also allows the user to submit the login form by pressing return.

Some notes:
- In the key-press handler I added some logic that auto focuses the username or password entry if they are empty. I wasn't sure whether this should go in `LoginModel::login`
  - I also think this should give some user feedback, i.e. logs in but password is empty, then password box is highlighted red and focused. But I don't know how to do this
- The key-press should probably be a method but I wasn't sure where was best to put it.

Fixes #70

### DIFF
--- a/src/app/components/login/login.rs
+++ b/src/app/components/login/login.rs
@@ -11,7 +11,7 @@ use crate::app::AppEvent;
 
 use super::LoginModel;
 
-const KEY_RETURN: u32 = 65293;
+const SUBMIT_KEY: &str = "Return";
 
 pub struct Login {
     dialog: gtk::Dialog,
@@ -25,20 +25,19 @@ fn handle_keypress(
     model: Rc<LoginModel>,
     event: &EventKey,
 ) -> Inhibit {
-    match event.get_keyval().to_glib() {
-        KEY_RETURN => {
-            let username_text = username.get_text().as_str().to_string();
-            let password_text = password.get_text().as_str().to_string();
-            if username_text.is_empty() {
-                username.grab_focus();
-            } else if password_text.is_empty() {
-                password.grab_focus();
-            } else {
-                model.login(username_text, password_text);
-            }
-            Inhibit(true)
+    if event.get_keyval().to_glib() == gdk::keyval_from_name(SUBMIT_KEY) {
+        let username_text = username.get_text().as_str().to_string();
+        let password_text = password.get_text().as_str().to_string();
+        if username_text.is_empty() {
+            username.grab_focus();
+        } else if password_text.is_empty() {
+            password.grab_focus();
+        } else {
+            model.login(username_text, password_text);
         }
-        _ => Inhibit(false),
+        Inhibit(true)
+    } else {
+        Inhibit(false)
     }
 }
 
--- a/src/app/components/login/login.rs
+++ b/src/app/components/login/login.rs
@@ -2,7 +2,7 @@ use gdk::EventKey;
 use gio::ApplicationExt;
 use glib::translate::ToGlib;
 use gtk::prelude::*;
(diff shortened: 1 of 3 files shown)


xou816:
(You really should set up rustfmt, it's fast and will save you a lot of trouble :p )

Thank you for the PR! This is a very welcome change!

src/app/components/login/login.rs
```diff
@@ -9,12 +11,37 @@ use crate::app::AppEvent;
 
 use super::LoginModel;
 
+const KEY_RETURN: u32 = 65293;
```
xou816, line 14: Have you tried using [keyval_from_name](https://gtk-rs.org/docs/gdk/fn.keyval_from_name.html)? Might handle the keycode conversion subtleties for us :)

src/app/components/login/login.rs
```diff
@@ -9,12 +11,37 @@ use crate::app::AppEvent;
...
+    username: gtk::Entry,
+    password: gtk::Entry,
+    model: Rc<LoginModel>,
+    event: &EventKey
+) -> Inhibit {
+    match event.get_keyval().to_glib() {
+        KEY_RETURN => {
+            let username_text = username.get_text().as_str().to_string();
+            let password_text = password.get_text().as_str().to_string();
+            if username_text.len() == 0 {
```
xou816, line 32: Prefer using [is_empty](https://doc.rust-lang.org/std/string/struct.String.html#method.is_empty) (same below)

src/app/components/login/login.rs
```diff
@@ -32,6 +59,16 @@ impl Login {
                 model.login(username, password);
```
xou816, line 59: The click handler could use the same handler and logic as below, if we rename it to something like `submit_login_form` or something, idk!

src/window.ui
```diff
@@ -349,7 +349,7 @@
                 <child>
                   <object class="GtkToggleButton" id="shuffle">
                     <property name="visible">True</property>
-                    <property name="can-focus">False</property>
+                    <property name="can-focus">True</property>
```
xou816, line 352: Is this intentional? Probably since can-focus makes sense here, but I'd rather have this in another PR (there are probably other focus/keyboard navigation issues in the main window too)

Douile: It was the only one I noticed in the main window so thought I would throw it in with these. I can remove it if you want though.

src/app/components/login/login.rs
Douile: I'll look into it

src/window.ui
xou816: Okay, why not :)

src/app/components/login/login.rs
Douile: @xou816 is 06df1a38df18457ab917f0c0182b150db72837f4 what you meant?

xou816:
Oh sorry, I should have answered you earlier about the Model thing: I'm keeping these Model structs mostly to manage interactions with the app state (reading from the state or sending "actions"). I'd actually prefer that submit fn to be on the Login struct :)

Douile:
no problem I will move it

xou816:
LGTM! :) I should make a release tomorrow, I'm glad that this will be part of it!

PR #80, opened 2021-02-19
Allow seekbar to grow when layout allows

### LINKED ISSUE #57: Add longer seekbar when layout allows
When the window is large the seekbar doesn't expand to fill, which for super long tracks makes seeking difficult.

image

It would be nice to have the seekbar expand to fill the available horizontal space for this.

Douile:
preview

Closes #57

### DIFF
--- a/src/window.ui
+++ b/src/window.ui
@@ -194,6 +194,8 @@
                   <object class="GtkBox">
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
+                    <property name="margin-start">10</property>
+                    <property name="margin-end">10</property>
                     <property name="orientation">vertical</property>
                     <child>
                       <object class="HdyClamp">
--- a/src/window.ui
+++ b/src/window.ui
@@ -191,40 +191,106 @@
                 <property name="can-focus">False</property>
                 <property name="hexpand">True</property>
                 <child type="center">
-                  <object class="HdyClamp">
+                  <object class="GtkBox">
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
-                    <property name="hexpand">True</property>
-                    <property name="maximum-size">320</property>
-                    <property name="tightening-threshold">200</property>
+                    <property name="orientation">vertical</property>
                     <child>
-                      <object class="GtkBox">
+                      <object class="HdyClamp">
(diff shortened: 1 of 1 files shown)


xou816:
You're on fire today!! Thanks again for your work!
It's very subjective of course, but I'm not very fond of this overly long slider! What do you think about clamping it to a reasonable size?

szbergeron:
@xou816 from a practicality standpoint I really like the length in the screenshot, since especially for podcasts and such fine tracking can get difficult. I'm ambivalent from an aesthetics standpoint but clamping length would probably reduce usability a bit at least.

szbergeron:
Would it potentially make sense to integrate the seekbar with the dividing line between the bottom bar and the rest of the UI? Sort of like how the spotify app does when the song is collapsed at the bottom (not in player view). It would allow the longest possible seekbar length without taking up any extra room.

Douile:
@xou816 personally I prefer the longer seek bar for the better accuracy when seeking. Maybe we could make this toggle-able in a settings menu?

xou816:
Nah, you've all convinced me :) but how about some extra spacing in the main bottom box? a little bit more, so long as it still allows the window to shrink to 360px

@Douile there seems to be one useless box in the hierarchy (maybe this was there before, not sure!) 
Capture d’écran du 2021-02-19 20-03-36

@szbergeron that's an interesting idea! I don't know where we'd put the track duration display though?

szbergeron:
I'll spare you my bad gimp skills, but elapsed/total time could either go under the play button or could display on hover over the seekbar (maybe only display the seekbar handle on hover too?) Think how the youtube seekbar UI works, just with the hovered time instead of the video preview.

xou816:
I really like this idea! Might be worth trying out! Maybe not in this PR, but we can totally experiment with this!

xou816:
About the spacing: Here is why I think some spacing is necessary :
image

Douile:
Ye I see what you mean, I added some margins to avoid that, lmk what you think. I believe should expand the seek bar to the right where the space isn't being used.

[Video of resizing with margins](https://streamable.com/p4o1k0)

xou816:
This is better, thanks! Merging :)

PR #105, opened 2021-02-26
#52: Add playlist track pagination

### LINKED ISSUE #52: The playlist is short
Only 100 tracks are displayed in the playlist and can be listened to, the rest are unavailable

Douile:
Add support for playlist track pagination, this allows loading of playlists with more than 100 songs.
Fixes #52


xou816:
Thank you for diving into this, looks great at first look!! :) Let me review that :)

src/app/backend/api.rs
```diff
@@ -491,7 +520,42 @@ impl SpotifyApiClient for CachedSpotifyClient {
...
+                    .await?;
+
+                let songs: Page<PlaylistTrack> = from_str(&song_text)?;
+
+                let mut songs: Vec<SongDescription> = songs.into();
+                let songs_loaded = songs.len();
+
+                tracks.append(&mut songs);
+
+                if songs_loaded < limit.try_into().unwrap() {
```
xou816, line 549: make limit a usize to avoid the conversion!

src/app/backend/api_models.rs
```diff
@@ -214,6 +214,17 @@ impl Into<Vec<SongDescription>> for DetailedPlaylist {
     }
 }
 
+impl Into<Vec<SongDescription>> for Page<PlaylistTrack> {
```
xou816, line 217: the api reference is really misleading, I just realized the generic `Tracks` used in this file in `DetailedPlaylist` is just a `Page` really, would you mind fixing that? this should allow  having only one conversion impl here (yours, for `Page<PlaylistTrack>`)

actually maybe we don't even need to have this tracks field at all on `DetailedPlaylist` since we no longer ask for the tracks field in `get_playlist_no_cache`!

Douile: I'm using Playlist as the first conversion as it doesn't have a tracks variable and then returning DetailedPlaylist as it has tracks so I can add the tracks to it. I'm not really sure what you mean by this.

xou816: Sorry, was thinking as I was typing! 

I re-took a look: the DetailedPlaylist type is no longer useful I believe, and can be removed!

xou816:
looks good to me :)

Issue #94, opened 2021-02-23
Playlists added in another Spotify app are not picked up


MartinWallgren:
**Describe the bug**
Playlists added in another spotify app are not picked up, not even after a restart of Spot.

**To Reproduce**
1. Go to 'the Playlist view in Spot'
2. Open a playlist that you are not yet following in another spotify app (for example web, mobile or anyone)
3. Follow the playlist
4. The playlists view in Spot is never updated.

**Expected behavior**
Expected the new playlist to appear in my Playlists view.

**Screenshots**
N/A

**General information:**
 - Distribution: Debian
 - Installation method [e. g. built from source, installed from Flathub...]: Flathub
 - Version [e.g. 0.1.0]:  0.1.8
 - Device used [e. g. desktop, phone...]: Desktop

**Stack trace:**
```
N/A
```

**Additional context**
I understand that Spot is in early development, but I think some mechanism to react to changes made outside Spot to the spotify account is needed at some point.

xou816:
Hi! Thanks for the report.

Playlists are picked up, but not immediately: the cache is not very smart and could use some improvments!

MartinWallgren:
Indeed, today the new playlists where there. Thank you for the reply.

xou816:
No problem! I'm keeping this open to remind myself to look into proper cache management

PR #108, opened 2021-02-27
Cache and api client improvements

### LINKED ISSUE #107: Cached responses in "wrong" directory
**Describe the bug**
Cached files are stored in the generic folder `~/.cache/net`.

**Expected behavior**
I think it would be better if cached files were stored under a spot directory. e.g. `~/.cache/spot/net`. As if someone is looking at their cached files they can better identify where they are from.

**General information:**
 - Distribution: N/A
 - Installation method [e. g. built from source, installed from Flathub...]: source
 - Version [e.g. 0.1.0]: master
 - Device used [e. g. desktop, phone...]: desktop

xou816:
should fix #107 among other things! also some clean up to make the api client easier to maintain


xou816:
and also #94

PR #109, opened 2021-02-28
Use return key constant instead of lookup

Douile:
Noticed I could use a constant here instead of a lookup.


xou816:
Oh even better indeed :)

Issue #96, opened 2021-02-23
[Feature] Add context menus


Douile:
Add context menus to items. e.g. playlist copy URL or spotify URI. This is aimed to make navigation easier. I have included lists of things I think each context menu should contain. This is definitely not essential so should be low priority imo. 

Songs:
- View artist(s) (note. songs and albums can have multiple artists)
- View album
- Copy link
- Copy URI

Albums: 
- View artist(s)
- Copy link
- Copy URI

Artists:
- Copy link
- Copy URI

Playlists:
- View owner
- Copy link
- Copy URI

xou816:
There are some context menus already actually, that can be added by any view that uses [`Playlist`](https://github.com/xou816/spot/blob/master/src/app/components/playlist/playlist.rs#L15), for instance the [Now playing](https://github.com/xou816/spot/blob/master/src/app/components/now_playing/now_playing_model.rs#L56) view :) I intend to add more contextual actions in the future but feel free to contribute some in the meantime!

Douile:
I have started working on this in https://github.com/Douile/spot/tree/patch/menus

Douile:
Would like it if these menus could be accessed by right clicking the row, I am looking into the best way to do this (for accessibility reasons should also support context menu key). 

Would also like to be able to right click playlist and album art (and maybe artist names) to be able to access similar context menus but this seems more complicated and I feel should be a separate PR to #111.

Douile:
Putting this here for my own reference: https://developer.gnome.org/gtkmm-tutorial/stable/sec-treeview-examples.html.en#treeview-popup-menu-example

@xou816 how do you feel about right click popup context menus?

xou816:
No objection, but probably to use actual old style menus but rather popovers :)

PR #111, opened 2021-02-28
#96: Add basic menus everywhere

Douile:
Adds basic menus to all places where Playlist is implemented. Not feature complete: Still can't add/remove songs from playlist, add songs to queue (queue not added yet).

References: #96

### DIFF
--- a/src/app/components/artist_details/artist_details_model.rs
+++ b/src/app/components/artist_details/artist_details_model.rs
@@ -146,10 +146,7 @@ impl PlaylistModel for ArtistDetailsModel {
         let copy_link = SimpleAction::new("copy_link", None);
         copy_link.connect_activate(move |_, _| {
             let clipboard = Clipboard::get(&SELECTION_CLIPBOARD);
-            clipboard.set_text(&format!(
-                "https://open.spotify.com/track/{}",
-                &track_id,
-            ));
+            clipboard.set_text(&format!("https://open.spotify.com/track/{}", &track_id,));
         });
         group.add_action(&copy_link);
 
--- a/src/app/components/artist_details/artist_details_model.rs
+++ b/src/app/components/artist_details/artist_details_model.rs
@@ -1,3 +1,5 @@
+use gio::prelude::*;
+use gio::{ActionMapExt, SimpleAction, SimpleActionGroup};
 use std::ops::Deref;
 use std::rc::Rc;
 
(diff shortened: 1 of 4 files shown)


Douile:
I have repeated a lot of code here, maybe I should make a util function to create the menus that are the same.

xou816:
Looks promising!! I'll review that asap, thanks!

Yeah, there's probably something we could do as a lot of menus will be similar, not sure how though! That's fine for now I suppose :)

xou816:
I'm curious, what kind of use do you see for the spotify URI? (spotify:track:etc)

src/app/components/artist_details/artist_details_model.rs
```diff
@@ -105,11 +109,68 @@ impl PlaylistModel for ArtistDetailsModel {
...
+                dispatcher.dispatch(AppAction::ViewArtist(id.clone()));
+            });
+            group.add_action(&view_artist);
+        }
+
+        let track_id = song.id.clone();
+        let copy_uri = SimpleAction::new("copy_uri", None);
+        copy_uri.connect_activate(move |_, _| {
+            let clipboard = Clipboard::get(&SELECTION_CLIPBOARD);
+            clipboard.set_text(&format!("spotify:track:{}", track_id.clone()));
```
xou816, line 141: not sure but track_id or &track_id might be enough for format, we only need to clone it once to move it into the closure

src/app/components/artist_details/artist_details_model.rs
```diff
@@ -105,11 +109,68 @@ impl PlaylistModel for ArtistDetailsModel {
...
-        None
+    fn menu_for(&self, id: String) -> Option<gio::MenuModel> {
+        let songs = self.tracks_ref()?;
+        let song = songs.iter().find(|song| song.id == id)?;
+
+        let menu = gio::Menu::new();
+        menu.insert(0, Some("View album"), Some("song.view_album"));
+        for (i, artist) in song.artists.iter().enumerate() {
+            menu.insert(
+                (i + 1) as i32,
```
xou816, line 167: I had not realized there was an append method, my bad :facepalm: would you mind using append here?

src/app/components/artist_details/artist_details_model.rs
Douile: > I'm curious, what kind of use do you see for the spotify URI? (spotify:track:etc)

In the official app if you search for a URI it will take you straight to that thing (playlist, artist, album, track): might be nice to add that here.

You can also use them to generate spotify codes (essentially QR codes you can scan in the app) https://www.spotifycodes.com/ (although it might be better to just let users copy codes directly)

Douile:
> I'm curious, what kind of use do you see for the spotify URI? (spotify:track:etc)

In the official app if you search for a URI it will take you straight to that thing (playlist, artist, album, track): might be nice to add that here.

You can also use them to generate spotify codes (essentially QR codes you can scan in the app) https://www.spotifycodes.com/ (although it might be better to just let users copy codes directly)

src/app/components/artist_details/artist_details_model.rs
Douile: Yep i think it would, my bad.

Douile:
> > I'm curious, what kind of use do you see for the spotify URI? (spotify:track:etc)
> 
> In the official app if you search for a URI it will take you straight to that thing (playlist, artist, album, track): might be nice to add that here.
> 
> You can also use them to generate spotify codes (essentially QR codes you can scan in the app) https://www.spotifycodes.com/ (although it might be better to just let users copy codes directly)

I do agree though they are less useful, so IDM if you want to remove that.

xou816:
I don't have a strong opinion about this, but if you're fine with it I'd like to keep it simple for now yeah :)

Douile:
Okay, fair. Remove both copy ones or just the URI and leave link?

xou816:
Link is fine! Seems useful to share tracks :)

xou816 (approved):

PR #121, opened 2021-03-02
Large seekbar above bottom controls

xou816:
As per @szbergeron suggestion, a seekbar above the bottom controls

I think I like how it looks, especially on mobile :) what do you think @Douile?

Capture d’écran du 2021-03-02 18-54-37

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -6,3 +6,21 @@
 .notification--shown {
   opacity: 1;
 }
+
+.seek-bar {
+  padding: 0;
+  padding-bottom: 2px;
+  min-height: 0;
+}
+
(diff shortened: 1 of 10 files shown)


Douile:
> As per @szbergeron suggestion, a seekbar above the bottom controls
> 
> I think I like how it looks, especially on mobile :) what do you think @Douile?
> 
> Capture d’écran du 2021-03-02 18-54-37

I like it

Douile:
@xou816 did you remove the shuffle button intentionally?

xou816:
Moved it to Now playing, I intend to add more queue control related controls here as well :)

Douile:
ahhh, okay that makes sense

PR #127, opened 2021-03-04
Typo in line 34

c-ortl:
recommandations -> recommendations


xou816:
Good catch, thanks :)

Issue #130, opened 2021-03-06
[feature request] Stream quality selection


coxackie:
Right now, one cannot select what stream quality to play the songs. Could this be please be implemented? (along with potentially other options?) I know that spotifyd allows this, so I am pretty sure it is doable with librespot intergration.

 Great work by the way. This is awesome and (almost too) lightweight.

xou816:
Hi! Thanks for the kind words :) This is not implemented in librespot, couldn't find any mention of it in their issue tracker either. No idea if this is feasible or not, but I can't help with that -- not knowledgeable about the inner workings of librespot!

coxackie:
hmmm... @xou816  from a quick google search:
https://github.com/librespot-org/librespot/wiki/Options
look at the line
```
Option | b | bitrate | Bitrate (96, 160 or 320). Defaults to 160 | BITRATE
```
So, presumably the sampling rate right now is sub-optimal... could you please have a better look, if possible?

xou816:
Ah, I misunderstood! Sure, I (or someone else) might have a look later :)

xou816:
Fixed in 0.1.10, can be changed with gsettings. Closing this for now, feel free to reopen if needed. I'll add an issue to implement a GUI for these settings.

PR #139, opened 2021-03-12
Change bitrate to 320

exciteabletom:
Solves #130 by using a 320kbps bitrate instead of the default 160kbps. All the other values are kept as the defaults.


xou816:
Hi! :) Thank you for taking the time to do this!

I'm sorry to tell you this, but I have a wip branch locally where I added a gsetting key to change the bitrate. I intended to leave 160 kbps as the default. I'll @mention you so you can have a look and give some feedback!

exciteabletom:
No worries at all, the gsetting sounds like a much better solution!

PR #140, opened 2021-03-12
GSettings

xou816:
@exciteabletom

see #139 and #130

### DIFF
--- a/data/dev.alextren.Spot.gschema.xml
+++ b/data/dev.alextren.Spot.gschema.xml
@@ -18,6 +18,18 @@
       <default>true</default>
       <summary>Prefer dark theme</summary>
     </key>
+    <key name="window-width" type="i">
+      <default>1080</default>
+      <summary>The width of the window</summary>
+    </key>
+    <key name="window-height" type="i">
+      <default>720</default>
+      <summary>The height of the window</summary>
+    </key>
+    <key name="window-is-maximized" type="b">
+      <default>false</default>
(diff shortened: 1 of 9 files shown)


exciteabletom:
All the options are working great on my machine, thanks for adding this :)

My only suggestion would be to have a GUI for changing the settings within Spot, but maybe that's a different PR entirely.

xou816:
Thanks! Glad it works as intended :)

Yeah, I'll add a GUI, most likely later indeed!

LennyPenny:
how do I use this? I tried to use the `gsettings` cli but I can't figure out what set of options to give it to change the bitrate for example. 

I also tried to use `dconf-editor` but the spot settings are nowhere to be found
I'm using the flatpak version btw

exciteabletom:
@LennyPenny 
Use `gsettings list-keys dev.alextren.Spot` to list all the options then use `gsettings set dev.alextren.Spot {option} {value}` to set an option.

So to set the bitrate to maximum would be `gsettings set dev.alextren.Spot player-bitrate 320`

xou816:
@exciteabletom thanks for the clarification :)

Also if you're using the flatpak'd version @LennyPenny you'll have to run the above commands from a sandboxed shell:

```
flatpak run --command=sh dev.alextren.Spot
gsettings list-keys dev.alextren.Spot
```

LennyPenny:
aah ok thx it was not clear to me that it works differently in the flatpak version
(maybe its worth mentioning that in the readme)

PR #144, opened 2021-03-14
Take buildtype into account when testing

dpeukert:
This prevents cargo from recompiling when running tests on a release build.


xou816:
Looks good to me! :) thanks for the fix -- and thanks for maintaining the AUR package!

PR #152, opened 2021-03-16
Github Actions: Add shellcheck

daPhipz:
This PR adds [shellcheck](https://github.com/koalaman/shellcheck) to the automatic testing on push.
There were almost no problems with the `sh` files, but I think it's better safe than sorry, isn't it? :wink:


xou816:
I hope I won't have to maintain more bash files than those two :D 

but alright, might be useful! :)

PR #168, opened 2021-03-18
Add Czech translation

ondras12345:
I tested it by building Spot with `gnome-builder`, it seems to work just fine.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,1 +1,1 @@
-en fr nl de
+en fr nl de cs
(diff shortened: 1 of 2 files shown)


xou816:
Thank you for your contribution! :)

daPhipz:
@ondras12345 Just out of curiosity - why is there a 3rd `msgstr` for "Songs selected"?

ondras12345:
Because Czech needs 3 plurals
https://github.com/xou816/spot/blob/8362d56234f84369fb4e2646a8dec3e00b0ee10d/po/cs.po#L18

https://github.com/xou816/spot/blob/8362d56234f84369fb4e2646a8dec3e00b0ee10d/po/cs.po#L98-L102

PR #181, opened 2021-03-25
Update spanish translation

alosarjos:
Use "Biblioteca" for "library", in the same way as the official Spotify client does.


xou816:
Trusting you on that :) thanks, merging!

PR #188, opened 2021-04-10
Add more mpris support

Douile:
Report all the information from SongDescription to MPRIS, this meant adding album name and art url.

I would've added more but it seems unecessary to add extra information to SongDescription
just for mpris reporting

### DIFF
--- a/src/dbus/mpris.rs
+++ b/src/dbus/mpris.rs
@@ -86,14 +86,7 @@ impl SpotMprisPlayer {
     }
 
     pub fn open_uri(&self, Uri: &str) -> Result<()> {
-        println!("Open Uri={}", Uri);
-        if let Some(id) = Uri.strip_prefix("/dev/alextren/Spot/Track/") {
-            self.sender
-                .unbounded_send(PlaybackAction::Load(id.to_string()).into())
-                .map_err(|_| Error::Failed("Coult not send action".to_string()))
-        } else {
-            Err(Error::NotSupported("Unsupported URI".to_string()))
-        }
+        Err(Error::NotSupported("Not implemented".to_string()))
     }
 
     pub fn pause(&self) -> Result<()> {
--- a/src/dbus/mpris.rs
+++ b/src/dbus/mpris.rs
@@ -146,7 +146,7 @@ impl SpotMprisPlayer {
     }
 
     pub fn seek(&self, Offset: i64) -> Result<()> {
-        if !self.can_seek() {
+        if !self.state.current_track().is_some() {
             return Ok(());
         }
 
@@ -164,7 +164,7 @@ impl SpotMprisPlayer {
     }
 
     pub fn set_position(&self, TrackId: ObjectPath, Position: i64) -> Result<()> {
-        if !self.can_seek() {
+        if !self.state.current_track().is_some() {
             return Ok(());
         }
 
--- a/src/dbus/mpris.rs
+++ b/src/dbus/mpris.rs
@@ -2,10 +2,11 @@
 #![allow(unused_variables)]
 
 use futures::channel::mpsc::UnboundedSender;
-use std::convert::Into;
+use std::convert::{Into, TryInto};
 use zbus::dbus_interface;
 use zbus::fdo::{Error, Result};
 use zbus::ObjectServer;
+use zvariant::ObjectPath;
 
 use super::types::*;
 use crate::app::{state::PlaybackAction, AppAction};
@@ -85,7 +86,14 @@ impl SpotMprisPlayer {
     }
 
     pub fn open_uri(&self, Uri: &str) -> Result<()> {
-        Err(Error::NotSupported("Not implemented".to_string()))
+        println!("Open Uri={}", Uri);
+        if let Some(id) = Uri.strip_prefix("/dev/alextren/Spot/Track/") {
+            self.sender
+                .unbounded_send(PlaybackAction::Load(id.to_string()).into())
+                .map_err(|_| Error::Failed("Coult not send action".to_string()))
+        } else {
+            Err(Error::NotSupported("Unsupported URI".to_string()))
+        }
     }
 
     pub fn pause(&self) -> Result<()> {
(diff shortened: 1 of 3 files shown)


xou816:
Oooh, nice!! I always assumed it would be more complicated to handle album art, that a remote URL wouldn't work

...actually doesn't show up for me in GNOME 40, is it working for you? Am I missing something?

Douile:
What you using to look at the metadata? It shows up for me in playerctl and URL leads to image. Maybe gnome doesn't like the metadata name I used from freedesktops spec https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/

To make a local image doesn't seem that hard, Could just request a temporary file name in /tmp and always write the image there when it updates (would be more complicated if we have to change name). That probably wouldn't on flatpak though.

The easiest way would be sending a base64 encoded dataURL instead imo, but idk if that's suported.

Douile:
**TODO**:
- [ ] OpenURI doesn't update UI
  - [ ] Now playing doesn't change
  - [ ] Status bar reports, no song playing
  - [ ] Seek bar dissapears
- [ ] OpenURI should maybe support playing playlists and albums
- [ ] OpenURI should maybe also support spotify URIs (because why not)
- [ ] Possibly implement shuffle
  - [ ] Current player state only supports toggling
    - [ ] Need to be able to explicitly set shuffle on/off
    - [ ] Need to be able to query shuffle state

xou816:
I see the metadata just fine, but I don't know how well mpris clients support remote urls (not great apparently).

That's why I've been putting this off for a while, it poses a few challenges: the album art must be downloaded ahead of time, and we must figure out a path outside the flatpak sandbox where to put the image. I had not though of /tmp actually -- maybe that would solve this second issue!

xou816:
Nice job on OpenURI! Looks interesting! I am not familiar with how clients use it however; how do they obtain the Uris in the first place?

As for your issues with the UI not updating: the Load action only sets some track as playing, but the track has to be in queue first

Douile:
Ye, noticed that. Struggling to find where I can stash a dispatcher to fetch the track data easily though.

Douile:
They can obtain the URI from the metadata, I'm planning on making it support the spotify URIs as well then people can just copy paste a spotify URI and play it in spot.

Douile:
There is also the TrackList reporting thing but that looks a pain to implement

xou816:
Oh right, it's a plain sender right now, it should be fine to replace that with a dispatcher instead!

Douile:
@xou816 sorry haven't had much time to work on this recently. I'm pretty new to rust and even newer to multithreaded rust :). I'm having problems using an `ActionDispatcher` instead of an `UnboundedSender`. I tried swapping them out but I can't get it to compile when it gets passed through threads. 

I've added `dispatcher: Box<ActionDispatcherImpl>` (using `Box<dyn ActionDispatcher>` wasn't working because it doesn't implement Clone but it's a trait :man_shrugging: ) to `SpotMpris` but when I get to 
https://github.com/Douile/spot/blob/649c8855fa7c830cd8351e4e6fa7348fed07ed18/src/dbus/mod.rs#L161-L168
It fails as I can't send between threads, I believe it's something todo with an UnboundedSender within the Worker reference in ActionDispatcherImpl. But I'm stuck, any chance you could bestow some rust wisdom upon me.

Thanks!

Douile:
Oh, I figured out the clone issue I think by adding 
```rust
impl Clone for Box<dyn ActionDispatcher> {
    fn clone(&self) -> Box<dyn ActionDispatcher> {
        self.box_clone()
    }
}
```
But I still can't send `dyn ActionDispatcher + 'static` between threads

xou816:
nw @Douile , I too had forgotten that PR, sorry about that!

Maybe for a start we could split this PR in two parts? The first few commits (for album art, seeking maybe? etc) should be fine to be merged :)

The rest is indeed a bit more complicated :grin: indeed our replacement for the plain sender has to be Send, I had not anticipated as much. 
The issue is that right now, the implementation is not Send, and the box_clone thing does not keep the Send trait either. To make it Send, we'd have to distinguish between a Send and a not Send version of the Worker (the latter allowing to execute non-Send Future tasks, for instance a task that affects a gtk widget directly).

One possible way would be to move the dispatch_local_async requirement to another trait. Then there'd be some more substantial work on having another implementation of this updated ActionDispatcher not depend on the current Worker thing, but on something similar that would be Send.

I'd gladly help you to look into this! :) if you go ahead with splitting the MR, I'll work against your branch on this part!

Douile:
Ok thanks, I'll revert to before adding the API requests. I'll also open an issue with the TODO list.

Douile:
[OpenURI](https://github.com/xou816/spot/pull/188/commits/5e9c82b916ef971bf2b81e919a9fc42a5437aa1d) introduces a bug where if you open a song not in the queue the UI will not update

xou816:
Yup, remove OpenURI for now and we'll work on it in a separate MR :) 

also would you mind squashing your commits to tidy up the history a bit? Then I'll review this first MR!

src/dbus/mpris.rs
```diff
@@ -145,11 +146,44 @@ impl SpotMprisPlayer {
     }
 
     pub fn seek(&self, Offset: i64) -> Result<()> {
-        Ok(())
+        if !self.can_seek() {
```
xou816, line 149: I'd prefer if calls to can_seek were replaced with `self.state.current_track().is_some()` or a dedicated non-pub fn, since can_seek is the public interface used by dbus clients, and the macros at play here could have unintended side effects!

src/dbus/mpris.rs
```diff
@@ -145,11 +146,44 @@ impl SpotMprisPlayer {
...
+    pub fn set_position(&self, TrackId: ObjectPath, Position: i64) -> Result<()> {
+        if !self.can_seek() {
+            return Ok(());
+        }
+
+        if TrackId.to_string() != self.metadata().id {
+            return Ok(());
+        }
+
+        // FIXME: Implicit position < 0 check by casting to u32
```
xou816, line 175: not sure what you would want to do? I think the TryInto or TryFrom implementation would return an Err variant if we had a position < 0?

src/dbus/mpris.rs
```diff
@@ -145,11 +146,44 @@ impl SpotMprisPlayer {
...
+        let pos: u32 = (Position / 1000)
+            .try_into()
+            .map_err(|_| zbus::fdo::Error::Failed("Could not parse position".to_string()))?;
+
+        if pos as u64 > self.metadata().length / 1000 {
```
xou816, line 180: Maybe compare Position and metadata().length earlier on? To avoid this division (and maybe a cast too?)

src/dbus/mpris.rs
Douile: Yes I must've deleted the commit that removed this FIXME when rebasing

src/dbus/mpris.rs
Douile: Unfortunately I think a cast is necessary here, but moving can avoid a division.

xou816:
Sorry should have mentioned that earlier but there's one other occurrence of can_seek that can be replaced!

Rest looks fine to me :)

Douile:
My bad should've seen this

xou816 (approved):

xou816:
No worries, merging :) thanks for your work!!

xou816:
hey  @Douile , I realize I have not taken proper time to test this; is the position/offset control working for you? Been playing with playerctl, and it's not :/ (getting a warning about trackId not being properly exposed as a dbus identifier, not sure if that's the issue or if it's an unrelated issue). Should I open an issue or am I missing something?

Douile:
> hey @Douile , I realize I have not taken proper time to test this; is the position/offset control working for you? Been playing with playerctl, and it's not :/ (getting a warning about trackId not being properly exposed as a dbus identifier, not sure if that's the issue or if it's an unrelated issue). Should I open an issue or am I missing something?

Set position works for me when I was testing, with something like `playerctl position 10`.

The trackId warning i think is fine, it's because the report the track id as a string and not as a dbus object. I looked into reporting as an object but I couldn't find an easy way to do it with our dbus library.

Douile:
I think there is an issue with seeking backwards actually, It throws could not parse position. Seeking forwards work. But defiantly worth opening an issue.

Douile:
That's cos I cast the offset to i32 :man_facepalming:

xou816:
Sorry, I just got confused by playerctl's postfix syntax for relative seeking!

But you're right about negative seeking! Good catch!!

Also no worries AT ALL, it's not released, no big deal, and well, it wasn't even working before :)

PR #195, opened 2021-04-22
CI: bump flatpak action to v3

bilelmoussaoui:
this also bumps the used image to gnome 40 to avoid re-downloading the runtime

### DIFF
--- a/.github/workflows/spot-snapshots.yml
+++ b/.github/workflows/spot-snapshots.yml
@@ -10,11 +10,12 @@ jobs:
     name: "Flatpak Builder"
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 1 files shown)


xou816:
Hi, thanks for that! Completely forgot to update the runtime here :) 

Also I see you added run-tests, great, I actually didn't know flatpak had this option (I suppose it only works for some build systems?), let's see how it goes but I might be able to remove my other workflow and keep this one only!

bilelmoussaoui:
> Hi, thanks for that! Completely forgot to update the runtime here :) 
> 
> Also I see you added run-tests, great, I actually didn't know flatpak had this option (I suppose it only works for some build systems?), let's see how it goes but I might be able to remove my other workflow and keep this one only!

It only works for meson/cmake I believe

PR #206, opened 2021-05-03
Fix MPRIS seek command

Douile:
This PR contains a few fixes:
- Allows negative seeking (backwards)
- Makes the seek command follow spec
  - If seeking to < 0 round to 0
  - If seeking past end of song skip the song ( this already worked as quirk of internals, but added concrete implementation)
- Updates the UI if you seek with MPRIS while the song is paused

### DIFF
--- a/src/app/components/playback/playback_control.rs
+++ b/src/app/components/playback/playback_control.rs
@@ -213,7 +213,8 @@ impl EventListener for PlaybackControl {
                 self.update_playing();
                 self.update_current_info();
             }
-            AppEvent::PlaybackEvent(PlaybackEvent::SeekSynced(pos)) => {
+            AppEvent::PlaybackEvent(PlaybackEvent::SeekSynced(pos))
+            | AppEvent::PlaybackEvent(PlaybackEvent::TrackSeeked(pos)) => {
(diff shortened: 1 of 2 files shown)


Douile:
> can the u128 to i128 conversion really fail?
Technically it can but I don't think it ever will cos we store most stuff internally as u32 anyway.

> (sorry for all the conversions, do not hesitate to change some of the types if they don't feel right (in the state struct or others). I might not have it thought through!)
Yeah I think it might be better to use larger numbers maybe internally, but I feel that should be a different PR

xou816:
Sure, but I mean in general, wouldn't that be safe to have `foo as i128` instead of try from?

(also you can merge the two match patterns, but good catch there, indeed the UI might not have been updated immediately!)

Douile:
Yeah very true :) ty

Douile:
btw do you prefer if I fix errors like this with a rebase to keep commit log clean or adding a new commit to fix them to make changes more clear?

xou816:
I prefer rebase, but that's fine either way :)

xou816 (approved):

xou816:
Thanks for the quick fix! :)

Issue #202, opened 2021-04-26
clippy: fix from_over_into warnings


xou816:
Fix this warning: https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into 

For instance:
```
warning: an implementation of `From` is preferred since it gives you `Into<_>` for free where the reverse isn't true
   --> src/app/state/playback_state.rs:364:1
    |
364 | impl Into<AppAction> for PlaybackAction {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider to implement `From` instead
```

PR #213, opened 2021-05-10
Fix "from_over_into" clippy warnings

nn1ks:
Hi,

First of all, thank you for this awesome app :)

This PR replaces the `Into` impls with `From` impls to fix clippy warnings as described in the issue #202.

### DIFF
--- a/src/api/api_models.rs
+++ b/src/api/api_models.rs
@@ -226,17 +226,17 @@ pub struct RawSearchResults {
     pub artists: Option<Page<Artist>>,
 }
 
-impl Into<ArtistSummary> for Artist {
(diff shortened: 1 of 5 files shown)


nn1ks:
I also noticed some other clippy warnings that I could fix. Should I include them in this pull request or open new pull requests for each warning?

xou816:
Hi! :)

Thanks for the praise and your contribution! This is certainly not the most exciting task, I am grateful someone freed me of it :p 

Other fixes are welcome, but I think I remember some warnings being up somewhat debatable. It would be easier to have a different PR for these, so as not to block this current PR.

I'll review this one a bit later, should not take too long!

PR #214, opened 2021-05-11
Fix some clippy warnings

nn1ks:
The only remaining warnings now are [from_over_into](https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into) warnings produced by the `glib_wrapper` macro (which probably would have to be fixed upstream) and [module_inception](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception) warnings. I didn't fix the "module_inception" warnings because it seems like you added the submodules on purpose and I think the warning is mostly there to warn beginners to not use `mod foo { ... }` inside foo.rs.

I think it might also be worth to consider adding clippy to the CI so future changes don't introduce new warnings.

### DIFF
--- a/src/api/api_models.rs
+++ b/src/api/api_models.rs
@@ -380,8 +380,8 @@ impl Playlist {
             art,
             songs: tracks.into(),
             last_batch: Batch {
-                batch_size,
                 offset,
+                batch_size,
                 total,
             },
             owner: UserRef {
--- a/src/app/components/selection/selection_tools.rs
+++ b/src/app/components/selection/selection_tools.rs
@@ -70,7 +70,7 @@ pub trait SelectionToolsModel {
         let action = if all_selected {
             SelectionAction::Deselect(songs.iter().map(|s| &s.id).cloned().collect())
         } else {
-            SelectionAction::Select(songs.iter().cloned().collect())
+            SelectionAction::Select(songs.to_vec())
         };
         self.dispatcher().dispatch(action.into());
     }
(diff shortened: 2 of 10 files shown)


xou816:
Perfect, thanks a lot!! You might want to add yourself to the CONTRIBUTORS file btw :)

Agreed about the module inception, maybe some day I'll reconsider but it's definitely not what clippy warns against. 

If we're going to add a clippy run to the spot-quality workflow, we'll just add the relevant flag!

xou816:
More detailed backtrace for the glib issue:

```
warning: an implementation of `From` is preferred since it gives you `Into<_>` for free where the reverse isn't true
    --> /.../glib-0.10.3/src/object.rs:730:9
     |
719  |   macro_rules! glib_object_wrapper {
     |  _-
     | |_|
     | |
720  | |     (@generic_impl [$($attr:meta)*] $name:ident, $ffi_name:path, $ffi_class_name:path, $rust_class_name:path, @get_type $get_type_expr:ex...
721  | |         $(#[$attr])*
722  | |         // Always derive Hash/Ord (and below impl Debug, PartialEq, Eq, PartialOrd) for object
...    |
730  | |         impl Into<$crate::object::ObjectRef> for $name {
     | |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...    |
1199 | /         $crate::glib_object_wrapper!(@generic_impl [$($attr)*] $name, $ffi_name, $ffi_class_name, $rust_class_name,
1200 | |             @get_type $get_type_expr);
     | |______________________________________- in this macro invocation (#3)
...
1230 | |     };
1231 | | }
```

xou816:
Looks like it's fixed on master! https://github.com/gtk-rs/gtk-rs/blob/master/glib/src/object.rs#L621 hopefully it makes into a patch :)

PR #245, opened 2021-06-28
Translation Brazilian Portuguese

lucraraujo:


po/pt-br.po
```diff
@@ -0,0 +1,163 @@
+# Tradução para português brasileiro do spot.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
```
ondras12345, line 2:
```suggestion
# Copyright (C) 2021 THE spot'S COPYRIGHT HOLDER
```
I _think_ this should look like this, see `en.po`:
https://github.com/xou816/spot/blob/108e27abddb5dbbcde24dfde001d39f192622e6c/po/en.po#L2

ondras12345:
Great!
I will test this as soon as @xou816 approves running the `spot-snapshots` workflow.

po/pt-br.po
```diff
@@ -0,0 +1,163 @@
...
+#: src/app/components/selection/selection_heading.rs:74 src/window.ui.in:26
+msgid "No song selected"
+msgstr "Nenhuma música selecionada"
+
+#. translators: This is part of a larger text that says "<n> songs selected" when in selection mode. This text should be as short as possible.
+#: src/app/components/selection/selection_heading.rs:88
+msgid "song selected"
+msgid_plural "songs selected"
+msgstr[0] "música selecionada"
+msgstr[1] "Músicas selecionadas"
```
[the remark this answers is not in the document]
xou816: absolutely :) 

> This is part of a larger text that says "<n> songs selected" when in selection mode. This text should be as short as possible.

I should migrate this kind of label so that they work like msgid="Add to {}"

xou816:
Thank you lucraraujo and as always with translations, ondras! :) Workflow should run

ondras12345:
Screenshot from 2021-06-28 23-10-58
Seems to work just fine

po/LINGUAS
```diff
@@ -1 +1 @@
-en ru fr nl de ca cs pl es pt tr id
+en ru fr nl de ca cs pl es pt tr id pt-br
```
ondras12345, line 1:
```suggestion
en ru fr nl de ca cs pl es pt tr id pt_BR
```
I hope this will fix it.

ondras12345:
The file needs to be renamed from `po/pt-br.po` to `po/pt_BR.po`.

ondras12345:
Ok, now it works.

Screenshot from 2021-06-29 14-47-13

Screenshot from 2021-06-29 14-48-07

xou816:
Looks good thank you both! Just curious, should we explicitly modify the existing Portuguese translation to be pt_PT as well? I'll do a bit of reading and then we should be able to merge

xou816:
Seems like it's right to omit PT (country code) in that case: https://www.gnu.org/software/gettext/manual/gettext.html#Header-Entry

> In this PO file field, but not in locale names, ‘ll_CC’ combinations denoting a language’s main dialect are abbreviated as ‘ll’. For example, ‘de’ is equivalent to ‘de_DE’ (German as spoken in Germany), and ‘pt’ to ‘pt_PT’ (Portuguese as spoken in Portugal) in this context. 

But it's not very clear as for LINGUAS: https://www.gnu.org/software/gettext/manual/gettext.html#po_002fLINGUAS

lucraraujo:
I checked in a few projects that uses PO files for translation and they omit the country code "PT". So, the language field for the pt.po can (or should?) be only "pt".
Do you want me to make this change?

xou816:
It seems to be fine to be more specific too, so let's try as is! :)

Issue #82, opened 2021-02-19
Clicking on the notification area box does not open or focus the app


echosa:
**Describe the bug**
When running Spot (or any media player in GNOME), a "notification" box shows up in the notifications area that allows easy access and control of music from anywhere. Screenshot:
![image](https://user-images.githubusercontent.com/133786/108562789-05b87d80-72c6-11eb-8ea3-432753a707a4.png)

The controls in this box work fine: play, pause, next, and previous. However, if I click anywhere else in the box (not on the controls) as I would a "normal" notification, I expect Spot to open, raise, or take focus. This does not happen.

**To Reproduce**
1. Open Spot
2. Start playing a song
3. Focus another window and/or minimize Spot
4. Open the GNOME activities area
5. Click on the notification box for Spot (but not the playback controls)

**Expected behavior**
Spot should come forward and/or be focused.

**General information:**
 - Distribution: Pop!_OS 20.10
 - Installation method [e. g. built from source, installed from Flathub...]: Flatpak through Flathub
 - Version [e.g. 0.1.0]: 0.1.7
 - Device used [e. g. desktop, phone...]: laptop

echosa:
I just tried this with the actual Spotify client, and it also does not raise when clicked, so I thought this might be a GNOME issue. However, when I try with [Lollypop](https://wiki.gnome.org/Apps/Lollypop), it does work.

xou816:
It's definitely an issue with spot, I am not properly implementing the MPRIS specification (I've been lazy!), specifically this : https://specifications.freedesktop.org/mpris-spec/latest/Media_Player.html#Method:Raise

Thank you for the report! :)

xou816:
fixed by 97ec540812193a0891e43c93fe5df91b56173aeb :)

PR #253, opened 2021-07-17
implement mpris Raise properly

eladyn:
As mentioned in #82, clicking the MPRIS-bubble in the notification center doesn't raise "Spot" in every case. 

What I did, is similar to what [Lollypop does](https://gitlab.gnome.org/World/lollypop/-/blob/15e7dc899502c5ed1f8e50e8c2c9224a3236f8e2/lollypop/mpris.py#L196-197), with the difference that instead of just calling `get_current_event_time()` there, we just let [gtk handle that](https://gitlab.gnome.org/GNOME/gtk/-/blob/77829cf3edf50b30ffb07b35bed9adcd0ecec3d3/gtk/gtkwindow.c#L5252-5266).

Calling `org.mpris.MediaPlayer2.Raise` method now causes the window to show itself if it was hidden before and fire a `"Spot" is ready` notification. 

A sidenote: In GNOME Shell, clicking the MPRIS-bubble doesn't necessarily call `Raise` (see [here](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/733a5e1acb004a3bd03f8e90182bd7129b105670/js/ui/mpris.js#L148-161)), so the actual behaviour might vary. However, it should work in any case now.

I'll be happy to adapt any changes or try some other things if necessary, just tell me!

### DIFF
--- a/src/app/components/window/mod.rs
+++ b/src/app/components/window/mod.rs
@@ -126,7 +126,7 @@ impl MainWindow {
     }
 
     fn raise(&self) {
-        self.window.show();
+        self.window.present();
     }
 }
 


xou816:
Hi!

Thank you for this PR and for taking the time to dig into the details of what was happening, this was an interesting read! You're absolutely right about the GNOME specific part, and it so happens that Spot uses present() already on app activation so all is good on that side.

I have only tested this in GNOME by manually dispatching the Raise signal this time, and it seems just fine :) Merging once tests pass!

Issue #180, opened 2021-03-25
Optimize list item layout for mobile.


wandy-dev:
The list of albums is not the best layout for mobile. This takes a long time to scroll through and it would be great if we could have the layout a little more mobile friendly. The same would also apply for playlists.

I managed to get a better layout to work using glade. However I haven't been able to figure out how to transition back to the original layout using libhandy. I build APIs for a living so this isn't really my forte. I'd love to pair with someone on this so I can learn and contribute more in the future.

I love this project by the way. It looks great on my pinephone, and my laptop!

### Screenshots

![current](https://user-images.githubusercontent.com/20841347/112489298-b34efe80-8d54-11eb-886f-c1c57faf92c3.png)

![mobile-albums](https://user-images.githubusercontent.com/20841347/112489453-d4175400-8d54-11eb-8df0-fe81acc25446.png)

xou816:
I like it! :) If someone wants to take a stab at it, I'd be more than happy to give some pointers!

There was also another person interested in a more compact list view (#162 ), both concerns could be addressed, with compact mode being used below a certain width threshold.

Probably can be done with a Squeezer

wandy-dev:
I'm definitely interested in giving this a shot. As far as the implementation goes, would creating two separate .ui files for mobile and desktop, then using a squeezer to switch between the two using a threshold be an acceptable solution? Part of me feels like that would be duplicating code in too many places. But again, I don't really have much experience with this stuff so perhaps that is the best solution.

xou816:
There might some duplication, but not all that much I believe!

Both the classic album widget and this new compact widget could use the same AlbumModel.

Then the Squeezer and the ListBox could be added to library.ui, and they could be bound to the same ListStore, use the same ScrolledWindow for loading more content, etc... haven't had a proper look but it seems possible, definitely!

wandy-dev:
Sorry for the delay on this. I've opened #254 which should get the ball rolling, but this is a bit above my pay grade.

Issue #162, opened 2021-03-17
Compact mode for large libraries


junibegood:
The main reason I'm looking for an alternative to the official Spotify client is that I like to add a lot of artists / albums in my library. Browsing such a list is painful in the official client because of the giant images associated with each item. The maximum number I can display at once on my computer screen is 18, while I have something like 200 or 250 artists in my library !

If Spot could bring a compact list mode with no image (or only thumbnails) and maybe smaller fonts, it would be awesome for users with large libraries like me ! Anyway, glad to see someone working on a Linux alternative to the official client, thanks for your work so far. :)

This feature request will likely appear as a bug report. Sorry about that, I didn't find how to create a feature request... I hope you can change the type afterwards !

xou816:
Thank you for the kind words!

That's reasonable and I can see how that'd be useful :) not my priority at the moment however!

fitojb:
This is a general issue with GNOME’s Adwaita theme, whose widgets are all over-padded and oversized.

PR #254, opened 2021-07-20
add compact album view

wandy-dev:
In issue https://github.com/xou816/spot/issues/180 I raised the idea that the album layout should change when spot is on a mobile device to make better use of the screen real estate. This also applies to https://github.com/xou816/spot/issues/162

I wanted to give this issue a shot, but unfortunately as a back end focused engineer, I don't have the skills, or time to acquire the skills, to complete this.

I figured I'd push up the layout changes that I made so that someone else could use them as a starting place.

### Changes
- Adds album_compact.ui
- Changes in album.rs to make the size of the album art work better on mobile. (These values are hard-coded and should be changed based off of the width of the window)

If anyone does pick this up, I'd love to be tagged in your PR so I can hopefully learn a bit more and perhaps make more substantial contributions in the future. Thanks!

### DIFF
--- a/src/app/components/album/album.rs
+++ b/src/app/components/album/album.rs
@@ -19,7 +19,7 @@ struct AlbumWidget {
 impl AlbumWidget {
     pub fn new() -> Self {
         screen_add_css_provider(resource!("/components/album.css"));
-        Self::from_resource(resource!("/components/album.ui")).unwrap()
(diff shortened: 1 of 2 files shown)


xou816:
Hi! Thanks for taking the time to do this :) this might help to kickstart things indeed!

Issue #163, opened 2021-03-17
Opening spot triggers "New login to Spotify" email


klingtnet:
**Describe the bug**

Opening spot triggers a "New login to Spotify" email.

**To Reproduce**

1. Just open spot.

**Expected behavior**

It should reuse its session or follow what the official client does.

**General information:**

 - Distribution: Arch Linux kernel 5.11.6
 - Installation method: AUR
 - Version: `0.1.10`
 - Device used: desktop

Douile:
We could reduce the frequency of these emails by trying to use the token, instead of just the username and password every time spot is opened. I will look into it.

Relevant auto-login code:
- [Event triggerer](https://github.com/xou816/spot/blob/8f4186550ea12d43fe32e6136b64de887512c5ed/src/app/components/login/login_model.rs#L15-L23)
- [Player notifier](https://github.com/xou816/spot/blob/1d361b1c81514dc6b5e42846456859dbb1f326e9/src/app/components/player_notifier.rs#L38-L40)

This however might be impossible with librespot, I'm not entirely sure.

xou816:
Yes, that was my original intent when storing the token along with username/password but AFAICT it is not possible to authenticate with that token in librespot. But I might have missed something -- afterall it's librespot that gives me this token... hm...

xou816:
Oh and also -- the mails stop coming after a while, don't know why ;p

alosarjos:
> Yes, that was my original intent when storing the token along with username/password but AFAICT it is not possible to authenticate with that token in librespot. But I might have missed something -- afterall it's librespot that gives us this token... hm...

Hmmm, this is in the code for sure...

https://github.com/librespot-org/librespot/blob/master/protocol/proto/authentication.proto#L25

Though it doesn't say anything about how to use it on the docs....

https://github.com/librespot-org/librespot/blob/master/docs/authentication.md

graynk:
This still happens for me, I built from source from master a couple of days ago. Do I have to change something?

xou816:
@nn1ks's patch should make it less frequent, but it might still happen and I don't think there is much else we can do :/ (I had not received any mail in a long while but I just had one today after Spotify's asked me to confirm my email)

graynk:
Hmm. I'm switching between Spot and [Psst](https://github.com/jpochyla/psst), maybe that's the reason why I get them daily

xou816:
Actually maybe Psst handles login better, doesn't seem to trigger emails! :thinking:

[...]

PR #255, opened 2021-07-20
Use token for autologin

nn1ks:
With these changes the saved token is used for autologin. The benefit of authenticating this way is that the "New login to Spotify" emails don't get send (#163).

The problem of the current code is that authentication fails if the token is expired. Two solutions I could think of:

- Save the expiration time of the token in the [Credentials](https://github.com/xou816/spot/blob/80a50debb0044900dd29df51d2a651e49803b0fc/src/app/credentials.rs#L10) struct and use password authentication if the token is expired.
- Always try to authenticate via the token and fall back to password authentication in case token authentication fails.

What do you think is the best solution?

### DIFF
--- a/src/app/credentials.rs
+++ b/src/app/credentials.rs
@@ -1,18 +1,9 @@
 use secret_service::{EncryptionType, Error, SecretService};
-use serde::{Deserialize, Serialize};
-use serde_json::{from_str, to_string};
+use serde::{de::Error as _, ser::Error as _, Deserialize, Deserializer, Serialize, Serializer};
 use std::collections::HashMap;
-use std::str::from_utf8;
+use std::time::{Duration, SystemTime};
 
 static SPOT_ATTR: &str = "spot_credentials";
-
-#[derive(Deserialize, Serialize, Clone, Debug)]
-pub struct Credentials {
-    pub username: String,
-    pub password: String,
-    pub token: String,
-    pub country: String,
-}
 
 fn make_attributes() -> HashMap<&'static str, &'static str> {
     let mut attributes = HashMap::new();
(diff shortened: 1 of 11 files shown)


Zaedus:
I completely agree this would be a great addition. I would also like to add that the need to store the token in gnome-keyring would be needless because the official spotify app stores the token at `~/.config/spotify/prefs`. This means the hassle of setting up PAM with the keyring is avoided, and spot becomes something that just works out of the box without additional user setup.

nn1ks:
> This means the hassle of setting up PAM with the keyring is avoided

Wouldn't you still need the keyring for the username?

> I think you should save the expiration time because its only a little bit more data to store, and then you would be able to send one less request.

Yeah I agree that it's a better solution although a downside of this is that users would probably have to delete (or manually change) the old entry in the keyring when upgrading to the next version (unless we make the field for the expiration time optional).

Zaedus:
> Wouldn't you still need the keyring for the username?

You could store it in some kind of config file like spotfiy does in the prefs file as well.

> users would probably have to delete (or manually change) the old entry in the keyring when upgrading to the next version

I propose that maybe we keep the current keyring format where the username and password are stored as a backup, but to optionally require it so that people who aren't/don't want to use the keyring can continue to login every so often to refresh their token which can be stored in a config file.

But these are just ideas. I'm not a rust dev so I don't know how much time it takes to implement all of this, and I respect you taking the time to contribute.

nn1ks:
I don't think it makes sense to only store the token and not the password because the token is only valid for a short amount of time (I think it is 1 hour). So the config file would also have to store the password (e.g. a shell command that outputs the password so it is not in plaintext) but implementing that would be out of scope for this pull request.

Zaedus:
Sorry, I wasn't too clear about what I meant. The token and username would be stored in plain text, while any sensitive credentials (the password) would be stored in the keyring. The one catch about the keyring is that it shouldn't be strictly necessary to use. For example, the program could first use the token stored in plain text, and if that fails, attempt to get the login credential from the keyring, and if that also fails (either the keyring is locked or the login request fails), then ask for the login details from the user and attempt to store them in the keyring. If it doesn't fail, then attempt to refresh the token. 

This way its not strictly necessary to use the keyring which in my opinion is better because it doesn't force additional setup.

nn1ks:
In my opinion all the credentials should be saved in one place and the keyring should be the default because that's how Gnome apps usually do it and it is also safer because the username and token is not in plain text. If you want to use a config file instead of the keyring you could specify it (e.g. with a command line option; something like `spot --config .config/spot/config`). The config file would contain all credentials (username, password, token) and the keyring would not be used at all.

This way you can specify a password if you don't use a keyring and therefore don't have to re-enter your username and password if you have the app closed for longer than an hour or so.

But let's wait on @xou816's opinion. I'm happy to implement either one :)

Zaedus:
Fair enough. That's some pretty sound reasoning, and I'm with you on keeping it all in one place. Would love to hear with xou thinks too.

xou816:
Hi there! 

Thank you very much for your work @nn1ks! I had not realized we could authenticate that way, or was too lazy to implement it, not sure :p is that the same token as the one used to call the API? That's really convenient!

As for the points that were raised:
- storing the expiry date sounds fine, as an Option probably; this is not an issue, a Credentials with a None date could be considered as expired, and it would therefore fallback to password auth, which is what we want for users updating;
- I'd rather avoid storing credentials in a file and having to handle all potential security concerns ;) however it might be interesting to see what kind of data can be "imported" from the official app, so that users could be prompted to automatically login _if_ they want to. Still, in a flatpak'd situation, we wouldn't have access to this config file anyway.

(One final thing, I just saw there's a Cache system in recent versions of librespot, I can't see where it is being used on the read side at the moment, but this might be useful in the future :thinking: )

I'll review this but the implementation looks fine overall!! :)

nn1ks:
> is that the same token as the one used to call the API?

Yes. I had to add the "streaming" scope but otherwise it just worked.

 > I just saw there's a Cache system in recent versions of librespot

As far as I can tell, the cache writes the credentials to a file in plain text, so I don't think we should use it for credentials.

src/app/credentials.rs
```diff
@@ -1,51 +1,86 @@
 use secret_service::{EncryptionType, Error, SecretService};
-use serde::{Deserialize, Serialize};
-use serde_json::{from_str, to_string};
+use serde::{de::Error as _, ser::Error as _, Deserialize, Deserializer, Serialize, Serializer};
 use std::collections::HashMap;
-use std::str::from_utf8;
+use std::time::{Duration, SystemTime};
```
xou816, line 4: maybe move these imports to the relevant module below, or even move this module to a file (optional)?

src/app/credentials.rs
```diff
@@ -1,51 +1,86 @@
...
-    let service = SecretService::new(EncryptionType::Dh)?;
-    let collection = service.get_default_collection()?;
-    let result = collection.search_items(make_attributes())?;
+mod serde_unix_timestamp {
+    use super::*;
 
-    let item = result.get(0).ok_or(Error::NoResult)?.get_secret()?;
-    let raw = from_utf8(&item).unwrap().to_string();
-    let parsed = from_str(&raw).map_err(|_| Error::Parse)?;
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<SystemTime>, D::Error>
```
xou816, line 17: maybe serialization/de serialization should be implemented for SystemTime rather than Option<SystemTime>?

src/app/credentials.rs
```diff
@@ -1,51 +1,86 @@
...
-pub fn save_credentials(creds: Credentials) -> Result<(), Error> {
-    let service = SecretService::new(EncryptionType::Dh)?;
-    let collection = service.get_default_collection()?;
-    let encoded = to_string(&creds).unwrap();
+#[derive(Deserialize, Serialize, Clone, Debug)]
+pub struct Credentials {
+    pub username: String,
+    pub password: String,
+    pub token: String,
+    #[serde(with = "serde_unix_timestamp")]
```
xou816, line 50: ah, I get it now!

maybe we should directly implement the traits rather than having this module? on a wrapper struct

src/app/credentials.rs
```diff
@@ -1,51 +1,86 @@
...
+    }
 
-    collection.create_item(
-        "Spotify Credentials",
-        make_attributes(),
-        encoded.as_bytes(),
-        true,
-        "text/plain",
-    )?;
+    pub fn retrieve() -> Result<Self, Error> {
```
xou816, line 63: :+1:  nice, I like that you moved everything in the struct :)

xou816 (commented):
looks good overall, thanks again!!

src/app/credentials.rs
nn1ks: The traits `Serialize` and `Deserialize` can't be implemented directly since `SystemTime` already implements them. (The reason I wrote custom serialize/deserialize functions is because `SystemTime` normally also has a field for nano seconds that isn't relevant for us because the time returned by the Spotify API isn't that accurate).

The downside of using a wrapper struct is that it has to be used each time `Credentials` is constructed.

An alternative would be to move the serialize/deserialize functions out of the module and add `#[serde(serialize_with = "serialize_unix_timestamp", deserialize_with = "deserialize_unix_timestamp")]`. This way there isn't an extra module in the file and no wrapper type has to be used.

What do you prefer?

xou816: Oh, missed that serde had an implementation for that already but you're right! Sure, we can't pretend to have it down to the nanoseconds, but for practical purposes I feel this would be fine to use the default serializer/deserializer?

nn1ks: Yeah makes sense. Added a commit to use the default serializer/deserializer.

xou816:
Thanks, merging :)

PR #262, opened 2021-08-02
Seek bar improvments

TotalDarkness-NRF:
This focuses on improving the functionality of the seek bar.
- Remove sending to same position for clicking, scrolling, and using arrow keys
- Add Fix for when gtk setting "gtk-primary-button-warps-slider" is false [*1]
- Add ability to use arrows keys to go forward or backwards by 5 seconds

Hopefully this helps to improve the seekbar. 

[*1] When the gtksetting is false the click to send it to that position will not work and will send the slider to around the same place it already was. Now all users should be able to be able to click the seekbar and have it work, regardless of there gtk settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ See [this comment](https://github.com/xou816/spot/issues/92#issuecomment-8018525
 Spot can also be configured via `gsettings` if you want to change the audio backend, the song bitrate, etc. [A GUI is planned but not available yet.](https://github.com/xou816/spot/issues/142)
 
 ### Seek bar warping
-It is possible to click on the seek bar to navigate to that position in a song. If you are having issues with this not working you may have [gtk-primary-button-warps-slide](https://docs.gtk.org/gtk3/property.Settings.gtk-primary-button-warps-slider.html) set to false.
+It is possible to click on the seek bar to navigate to that position in a song. If you are having issues with this not working you may have [gtk-primary-button-warps-slider](https://docs.gtk.org/gtk3/property.Settings.gtk-primary-button-warps-slider.html) set to false.
 In order to fix this issue set the value to true in your gtk configuration.
 
 ### Scrobbling
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ See [this comment](https://github.com/xou816/spot/issues/92#issuecomment-8018525
 
 Spot can also be configured via `gsettings` if you want to change the audio backend, the song bitrate, etc. [A GUI is planned but not available yet.](https://github.com/xou816/spot/issues/142)
 
+### Seek bar warping
+It is possible to click on the seek bar to navigate to that position in a song. If you are having issues with this not working you may have [gtk-primary-button-warps-slide](https://docs.gtk.org/gtk3/property.Settings.gtk-primary-button-warps-slider.html) set to false.
+In order to fix this issue set the value to true in your gtk configuration.
+
 ### Scrobbling
 
 Scrobbling is not supported directly by Spot. However, you can use a tool such a [rescrobbled](https://github.com/InputUsername/rescrobbled) ([see #85](https://github.com/xou816/spot/issues/85)).
--- a/src/app/components/playback/playback_control.rs
+++ b/src/app/components/playback/playback_control.rs
@@ -121,11 +121,13 @@ impl PlaybackControl {
         widget.seek_bar.connect_change_value(
             clone!(@weak model, @weak track_position => @default-return signal::Inhibit(false), move |_, scroll, mut requested| {
                 match scroll {
-                    gtk::ScrollType::StepForward => {
+                    gtk::ScrollType::StepForward | gtk::ScrollType::PageForward => {
                         let duration = model.current_song_duration().unwrap_or(0.0);
                         requested = min_by(requested + STEP, duration, |a, b| a.partial_cmp(b).unwrap())
                     },
-                    gtk::ScrollType::StepBackward => requested = max_by(requested - STEP, 0.0, |a, b| a.partial_cmp(b).unwrap()),
+                    gtk::ScrollType::StepBackward | gtk::ScrollType::PageBackward => {
+                        requested = max_by(requested - STEP, 0.0, |a, b| a.partial_cmp(b).unwrap())
+                    },
                     _ => (),
                 }
                 track_position.set_text(&format_duration(requested));
--- a/src/app/components/playback/playback_control.rs
+++ b/src/app/components/playback/playback_control.rs
@@ -125,13 +125,12 @@ impl PlaybackControl {
 
         widget.seek_bar.connect_button_press_event(
             clone!(@weak model => @default-return signal::Inhibit(false), move|scale, event| {
-                let mut x = event.position().0;
+                let (mut x, y) = event.position();
                 let offset = scale.layout_offsets().0 as f64;
+                println!("y: {}", event.position().1);
                 if offset <= x { x-= offset; } // Just in case future has some offset
                 let width = scale.range_rect().width as f64;
-                // TODO figure out why sometimes clicking doesnt change song seconds
-                // Clicking slightly under seek changes the value but doesnt actually seek
-                if x >= 0.0 && width > 0.0 {
+                if y >= 0.0 && y <= 20.0 && x >= 0.0 && width > 0.0 {
                     scale.set_value(model.current_song_duration().unwrap_or(0.0) * (x / width));
                 }
                 signal::Inhibit(false)
--- a/src/app/components/playback/playback_control.rs
+++ b/src/app/components/playback/playback_control.rs
@@ -144,22 +144,6 @@ impl PlaybackControl {
             }),
         );
 
-        widget.seek_bar.connect_button_press_event(
-            clone!(@weak model => @default-return signal::Inhibit(false), move|seek_bar, event| {
-                let (x, y) = event.position();
-                let width = seek_bar.range_rect().width as f64;
-                if (3.0..=20.0).contains(&y) && width >= 16.0 && (9.0..=width - 9.0).contains(&x) {
-                    let x = if x > 16.0 { x - 16.0 } else { 0.0 };
-                    let amount = model.current_song_duration().unwrap_or(0.0) * (x / (width - 16.0));
-                    let amount = if amount < 3000.0 { amount } else { amount + 3000.0 };
-                    seek_bar.set_value(amount);
-                    signal::Inhibit(false)
-                } else {
-                    signal::Inhibit(true) // Ignore
-                }
-            }),
-        );
-
         widget.seek_bar.connect_scroll_event(move |_, _| {
             /*
             Scrolling doesnt work well. Only shows direction if the slider is clicked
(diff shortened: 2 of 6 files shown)


xou816:
Hi! thank you again for your work :) 

clicking is already part of the default behavior of the gtk Scale, I don't see why this is needed, could you elaborate?

TotalDarkness-NRF:
> clicking is already part of the default behavior of the gtk Scale, I don't see why this is needed, could you elaborate?

My understanding is that clicking on the slider (circle) and dragging it was the only way to move it. Every time I would attempt to click on a random spot of the seekbar it would not change to that spot. What I have done it made it change to the spot you have clicked on that seekbar without needing to drag the slider each time.

xou816:
Well then this sounds like a bug :/ the scale is definitely clickable and we should not mess with additional click handlers! Do you have similar issues in other applications using a gtk scale? Could this be coming from a custom theme, or from Spot making the scale too thin?

TotalDarkness-NRF:
Im not sure I can test  in vms to see if its just me. I do not have any other applications I can test to see.
The clicking on the scale work but what it does is odd. It will not go to where you click but go back less than a second. 
I have tried changing all themes back to default but that did not seem to change anything

TotalDarkness-NRF:
The issue seemed to be at my end.
In order to fix I edited ~/.config/gtk-3.0/settinings.ini and set gtk-primary-button-warps-slider=true

Is there a way to make this setting true for the application, or should we let users know that that should be enabled if they want to click and seek
For now I will remove uneeded code and keep the rest. I will also look into these gsettings to see what I can do

src/main.rs
```diff
@@ -81,7 +81,9 @@ fn startup(settings: &settings::SpotSettings) {
         .set_gtk_application_prefer_dark_theme(settings.prefers_dark_theme);
 
     //Make the slider move on click
-    gtk::Settings::default().unwrap().set_gtk_primary_button_warps_slider(true);
+    gtk::Settings::default()
```
TotalDarkness-NRF, line 84: Is this sufficient to make the slider work? Based on what I understand this will only change it for the application.

xou816:
I am surprised this isn't the default behavior but ultimately I think it's up to the user to pick which behavior they prefer? Will think it over!

But thanks, I'll have a look at the PR later :)

src/app/components/playback/playback_control.rs
```diff
@@ -115,18 +115,43 @@ impl PlaybackControl {
...
         let track_position = &widget.track_position;
+
         widget.seek_bar.connect_change_value(
-            clone!(@weak model, @weak track_position => @default-return signal::Inhibit(false), move |_, _, requested| {
+            clone!(@weak model, @weak track_position => @default-return signal::Inhibit(false), move |_, scroll, mut requested| {
+                let mut debounce = 200;
+                match scroll {
+                    gtk::ScrollType::StepForward => {
+                        let duration = model.current_song_duration().unwrap_or(0.0);
+                        requested = if requested + 5000.0 <= duration {
```
xou816, line 126: extract step value to a local variable or a constant

xou816: or simply min(requested + step, duration)

src/app/components/playback/playback_control.rs
```diff
@@ -115,18 +115,43 @@ impl PlaybackControl {
...
+                            requested + 5000.0
+                        } else { duration };
+                        debounce = 100;
+                    },
+                    gtk::ScrollType::StepBackward => {
+                        requested = if requested >= 5000.0 {
```
xou816, line 132: (similarly, max(requested - step, 0) )

src/app/components/playback/playback_control.rs
```diff
@@ -115,18 +115,43 @@ impl PlaybackControl {
...
-            clone!(@weak model, @weak track_position => @default-return signal::Inhibit(false), move |_, _, requested| {
+            clone!(@weak model, @weak track_position => @default-return signal::Inhibit(false), move |_, scroll, mut requested| {
+                let mut debounce = 200;
+                match scroll {
+                    gtk::ScrollType::StepForward => {
+                        let duration = model.current_song_duration().unwrap_or(0.0);
+                        requested = if requested + 5000.0 <= duration {
+                            requested + 5000.0
+                        } else { duration };
+                        debounce = 100;
```
xou816, line 129: i'm fine with having debounce set to 100 in all cases :)

src/app/components/playback/playback_control.rs
```diff
@@ -115,18 +115,43 @@ impl PlaybackControl {
...
+                    },
+                    gtk::ScrollType::StepBackward => {
+                        requested = if requested >= 5000.0 {
+                            requested - 5000.0
+                        } else { 0.0 };
+                        debounce = 100;
+                    },
+                    _ => (),
```
xou816, line 137: do you think we should handle other scroll types as well ?

src/app/components/playback/playback_control.rs
```diff
@@ -115,18 +115,43 @@ impl PlaybackControl {
...
+                debouncer_clone.debounce(debounce, move || {
                     model.seek_to(requested as u32);
                 });
                 signal::Inhibit(false)
             }),
         );
 
+        widget.seek_bar.connect_scroll_event(move |_, _| {
+            /*
+            Scrolling doesnt work well. Only shows direction if the slider is clicked
```
xou816, line 149: do you mind explaining a bit more, what direction are you referring to? is it actively stopping the rest from working?

src/main.rs
xou816: right, but i'd rather leave that to its default value so the slider acts consitently with other apps (and the user's preferences)

TotalDarkness-NRF: Alright fine but maybe in the readme include this just in case a user may not know that. I for a while had just assumed that it was not implemented.

src/app/components/playback/playback_control.rs
TotalDarkness-NRF: Yes I was also thinking to have step value be a setting in the future.

src/app/components/playback/playback_control.rs
TotalDarkness-NRF: I can look at this but most are already handled (jump, page foward, page backwards). The rest I can never find a way to activate them

src/app/components/playback/playback_control.rs
TotalDarkness-NRF: So when this event is triggered it will send EventScroll with the scroll direction included. Unfortuantley the only way this sends the proper up or down is when the user is clicking the slider knob while scrolling. Otherwise it is always sends as smooth.
The reason I decided to do this instead is to remove an issue where scrolling would not work. (similar to when clicking was not working but that was a setting issue)
Unless it does work and it is once again another setting issue.

src/main.rs
xou816: sure, go ahead :)

src/app/components/playback/playback_control.rs
xou816: as for the min max : have a look at https://doc.rust-lang.org/std/primitive.f64.html#method.max it should be much simpler :)

src/app/components/playback/playback_control.rs
xou816: ah interesting! so which event triggers the Step* ones? (sorry if I am adding a lot of comments to your PR and asking a lots fo questions, it is an area of gtk that I don't know much about!)

src/app/components/playback/playback_control.rs
xou816: so when it's sending a ScrollDirection::Smooth, it prevents scrolling to change the value?

because in my case, scrolling offsets the knob in the expected direction so all seems good! it's not incredibly convenient, but it's not broken :thinking:

src/app/components/playback/playback_control.rs
xou816: okay so wait I'm confused, what user action results in a StepForward/StepBackward event?

xou816: (might be interesitng to use https://gtk-rs.org/gtk3-rs/stable/latest/docs/gtk/prelude/trait.RangeExt.html#tymethod.set_increments however )

xou816: okay nevermind, a larger variety of event trigger the step ones, but the others not so much, they might be related to others implementations of Ranges!

xou816: but I can confirm that set_increments is sufficient, no need to reimplement the stepping logic i think :)

src/app/components/playback/playback_control.rs
TotalDarkness-NRF: Hmm this might be another configuration issue. For me it does the same thing it did with clicking (before I discoverd the config option).
What are the gtk config options for the flathub builds?

src/app/components/playback/playback_control.rs
TotalDarkness-NRF: Hmm that might work. I must have missed that. I knew the issue was that it was try to send the default page size which was probably 0 but didnt know how to change

src/app/components/playback/playback_control.rs
```diff
@@ -110,13 +110,14 @@ pub struct PlaybackControl {
...
 
+const STEP: f64 = 5000.0;
 impl PlaybackControl {
     pub fn new(model: PlaybackControlModel, widget: PlaybackControlWidget) -> Self {
         let model = Rc::new(model);
         let debouncer = Debouncer::new();
         let debouncer_clone = debouncer.clone();
-
         let track_position = &widget.track_position;
+        widget.seek_bar.set_increments(STEP, STEP);
```
xou816, line 120: I don't know the default values but I feel like the page size should be greater than the regular step size!

README.md
```diff
@@ -33,6 +33,10 @@ See [this comment](https://github.com/xou816/spot/issues/92#issuecomment-8018525
 
 Spot can also be configured via `gsettings` if you want to change the audio backend, the song bitrate, etc. [A GUI is planned but not available yet.](https://github.com/xou816/spot/issues/142)
 
+### Seek bar warping
+It is possible to click on the seek bar to navigate to that position in a song. If you are having issues with this not working you may have [gtk-primary-button-warps-slide](https://docs.gtk.org/gtk3/property.Settings.gtk-primary-button-warps-slider.html) set to false.
```
xou816, line 37: (small typo, slideR)

src/app/components/playback/playback_control.rs
TotalDarkness-NRF: The defaults are both 0. From documentation:
 "The step size is used when the user clicks the GtkScrollbar arrows or moves GtkScale via arrow keys. The page size is used for example when moving via Page Up or Page Down keys"
I think in this case it works out fine to have it at the same value. Maybe in the future have it as a setting that users can change?

src/app/state/app_state.rs
```diff
@@ -77,7 +77,7 @@ impl AppState {
         }
     }
 
-    pub fn recommanded_context(&self) -> SelectionContext {
+    pub fn recommended_context(&self) -> SelectionContext {
```
xou816, line 80: good catch :)

src/app/components/playback/playback_control.rs
xou816: fair enough! :)

xou816 (approved):

xou816:
Thank you! :)

PR #270, opened 2021-08-10
Mpris repeat playback controls fix

TotalDarkness-NRF:
Add simple fix for playback controls not going back or forward when it should while in repeat mode for mpris
Now expected behaviour will allow playback to go back or forward whenever while repeating, (same behaviour as the normal spot player)

### DIFF
--- a/src/app/state/playback_state.rs
+++ b/src/app/state/playback_state.rs
@@ -83,6 +83,7 @@ pub struct PlaybackState {
     position: Option<Position>,
     pub source: Option<PlaylistSource>,
     current_batch: Option<Batch>,
+    repeat: RepeatMode,
     is_playing: bool,
 }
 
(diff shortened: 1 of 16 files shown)


xou816:
Haha I can barely keep up with all your PR! 

This is nice! But this additional field shouldn't be needed, `has_prev_next` should return the correct result (or else it's a bug), it might need to be refreshed when the event is received however

TotalDarkness-NRF:
>`has_prev_next` should return the correct result (or else it's a bug), it might need to be refreshed when the event is received however
Yes that appeared to be the bug. I was unaware that there was even the prev and next functions, which did not follow the proper repeat indexes. It should be following it now

src/app/state/playback_state.rs
```diff
@@ -314,36 +311,39 @@ impl PlaybackState {
         self.is_playing = false;
     }
 
+    fn play_index(&mut self, index: Option<usize>) -> Option<String> {
```
xou816, line 314: nitpick: callers of play_index should have to handle the Option and unwrap it somehow, since there's no point in calling play_index(None)

src/app/state/playback_state.rs
```diff
@@ -139,17 +139,14 @@ impl PlaybackState {
     }
 
     pub fn prev_song(&self) -> Option<&SongDescription> {
-        self.position
-            .filter(|pos| pos.index > 0)
-            .map(|pos| &self.running_order()[pos.index - 1])
+        self.prev_index()
```
xou816, line 142: nice! :+1:

src/app/state/playback_state.rs
xou816: good idea otherwise to have this function!

xou816:
Looks good to me thanks a lot for the fix :)

PR #275, opened 2021-08-20
Update dependencies

sigaloid:
https://pastebin.com/raw/7f8QHPVg

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,9 +47,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.15"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7404febffaa47dac81aa44dba71523c9d069b1bdc50a77db41195149e17f68e5"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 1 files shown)


sigaloid:
Odd. Running tests after reverting the test.

sigaloid:
Double odd. Passes tests when I revert. Updating Isahc to 1.4.1 seems to trigger it according to the actions log.

sigaloid:
Ok [cargo update](https://lib.rs/crates/cargo-update) does not feel like cooperating and mangling with the Cargo.lock still doesn't pass tests. Plus it seems like you have a dependency bot, so for now closing.

xou816:
Hmm, interesting, I thought dependabot would take care of these but actually not! Might be worth digging into it, I tried it locally and there were compilation errors -- problems for later :D

xou816:
zbus seems to break a few things (expected, 2.x is in beta), the rest should be safe

I should also change the libadwaita dependency to allow newer versions as well

sigaloid:
Hm, should I attempt undoing that dependency in cargo.lock? I wish cargo update had a --exclude option, instead of manually fudging around with Cargo.lock which is hardly good practice AFAIK.

xou816:
As you want, no emergencies here :) but yeah agreed wish there was that!

xou816:
Ah and if you do -- you might have to run `ninja cargo-sources.json -C target` so that the cargo-sources.json file (which is used to prefetch the crates) is up to do date!

PR #293, opened 2021-08-29
Adding Purism's custom FormFactor specification

1peter10:
This should improve discoverability of the app in PureOS store on the Librem 5 and help with Phosh on other (e.g. PinePhone) distributions.


xou816:
Interesting, thanks! I haven't tested it in a VM in a while, hope it still deserves to be said to support this form factor :p

PR #328, opened 2021-09-23
Added spotify URI handler

bbb651:
This is still very WIP, I managed to register the `spotify` URI.
I ran into a problem with ownership that I'm not sure how to tackle, when the `open` signal is called the `activate` signal isn't, so if in the `open` closure we need to open/focus the window. The problem is the closure in `Application::connect_activate` and closure in `Application::connect_open` both need to take ownership of `window` and `sender` because they have static lifetimes and the compiler thinks they might be able to outlive `main` (even though they can't).

### DIFF
--- a/data/dev.alextren.Spot.desktop
+++ b/data/dev.alextren.Spot.desktop
@@ -1,10 +1,11 @@
 [Desktop Entry]
 Version=1.0
 Name=Spot
-Exec=spot
+Exec=spot %u
 Icon=dev.alextren.Spot
 Terminal=false
 Type=Application
 Categories=GTK;GNOME;Music;AudioVideo;
+MimeType=x-scheme-handler/spotify;
 StartupNotify=true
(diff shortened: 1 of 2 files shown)


xou816:
Nice work!!

The easiest way to proceed is to clone the `sender` and use it to dispatch a new action (`AppAction::OpenURI(..)` perhaps) that in turn leads to some `AppEvent::OpenedURI` event. This event can then be handled by any `EventListener`, and more specifically by the [`window`](https://github.com/xou816/spot/blob/master/src/app/components/window/mod.rs) component (which holds a reference to the actual Gtk window).

You might need another `EventListener` to dispatch the needed actions.

xou816:
initial support for this in added in 917fdee2aa9e2bb72fab5a4b459670eadbb0b618 closing this for now! thanks a lot @bbb651 :)

xou816:
closing this for now

PR #345, opened 2021-10-14
Added info about Caps Lock in login screen

przebor:
Hello 👋 
I saw that there's no info about "Caps Lock being enabled" on Spot login screen, so...
I implemented it.

Please let me know if something's wrong with my code (I hope not haha)

### DIFF
--- a/po/spot.pot
+++ b/po/spot.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: spot\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-10-02 16:37-0400\n"
+"POT-Creation-Date: 2021-10-16 13:23+0200\n"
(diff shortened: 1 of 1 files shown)


xou816:
Hi! This looks absolutely fine, thank you for your contribution!! :) I merely have one question atm: is the explicit dependency on gdk4-sys needed? can we work around that by updating our version of gdk4?

przebor:
Ohh i just realized we don't need that.
Is was using gdk4-sys to check caps lock state but it was unsafe so I did it other way and forgot about adding it there.
When I get back to my PC, I will remove the dependency.

przebor:
Removed `gdk4-sys` from `Cargo.toml` and `cargo-sources.json` and squashed commits.
Should be fine now.

przebor:
Should be fine now :)

xou816:
Perfect, thanks!! merging this :)

PR #346, opened 2021-10-21
Flatpak: Use master SDK, don't build libadwaita

jannuary:
GNOME Platform has libadwaita as is, so there's no need to build it separately.

Currently broken:

- [x] Dark style preference 
- [x] Add in-app setting for dark mode
- [x] New icons too big:
image
- [x] Album grid layout busted: 
image
- [ ] Fix CI again


xou816:
I don't know how we'll handle merging the flatpak manifests yet, but I'm going to make a development branch -- wasn't really needed until now but looks like it might be useful after all :) I'll give you a heads up when it's there so you can rebase your PR against it

PR #358, opened 2021-11-03
Draft: Add animated playback indicator

jannuary:
Adds an animated playback indicator to indicate playback in an animated fashion:

Kooha-11-03-2021-17-18-16

Credit goes to Danielle Foré for implementing this in her [elementary Music branch](https://github.com/elementary/music/tree/playback-indicator).

TODO:

- [x] Figure out how to actually pause the animation while playback is paused
- [ ] Figure out why the playback state is always `false` when reopening album


xou816:
Learning from the best <3 looking very good so far! to pause the animation, I think toggling a class that sets [`animation-play-state`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-play-state) should work

xou816:
just realized you wrote this:

> Figure out why the playback state is always false when reopening album

when you reopen an album, the widget is recreated, and thus it's too late to listen on incoming events -- you need to pull the info from the state.

the state needs to be read and set when first creating each songmodel, [like it is here](https://github.com/xou816/spot/blob/8bf0a62ca292da0bd8bdfc06c00f9e46d0be95c7/src/app/components/playlist/playlist.rs#L100)

which brings me to my next point -- I think it'd be nice to handle this new property (paused or not) in songmodel just like the others

I'd be glad to help with the technical details if needed!

jannuary:
Won't be able to ship this, sorry! Feel free to pick up if anyone wants though

xou816:
No worries!! Thank you for your work!!

PR #363, opened 2021-11-08
Automatically try other ports in case of timeouts

### LINKED ISSUE #343: Try multiple AP ports in case of timeouts
I have created an issue #337, so that users can select which port to use in restrictive networks. And now I'm thinking it might be better to automatically try another access point after (shorter) timeout due to blocked ports. Spotify makes use of port 4070 which is not so well-known.

A simple solution would be to create a loop over known port numbers (80, 443 and 4070) and to call Session::connect with each port number. The successful ones can be stored in gsettings to skip unsuccessful ones at the future logins (or we might not have to go as far as to do so since there is no warranty that those ports are always opened, and users are not expected to log in so often).

sei0o:
Fixes #343. `SessionError::IoError` corresponds to a timeout. If `ap-port` value in gsettings is set, it won't connect to any other port.

### DIFF
--- a/src/player/player.rs
+++ b/src/player/player.rs
@@ -223,7 +223,7 @@ playlist-modify-public,\
 playlist-modify-private,\
 streaming";
 
-const KNOWN_AP_PORTS: &[u16] = &[80, 443, 4070];
+const KNOWN_AP_PORTS: [Option<u16>; 4] = [None, Some(80), Some(443), Some(4070)];
 
 async fn get_access_token_and_expiry_time(
     session: &Session,
@@ -253,13 +253,9 @@ async fn create_session(
                 dbg!(e);
                 SpotifyError::LoginFailed
             })
-        },
+        }
         None => {
-            let mut ret = None;
-
-            let mut ports: Vec<Option<u16>> = KNOWN_AP_PORTS.iter().map(|port| Some(port.to_owned())).collect();
-            ports.push(None);
-            for port in ports {
+            for port in KNOWN_AP_PORTS {
                 let session_config = SessionConfig {
                     ap_port: port,
                     ..Default::default()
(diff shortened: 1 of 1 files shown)


sei0o:
rustfmt says `error[E0670]: `async fn` is not permitted in Rust 2015` ...Seems strange, will look into it later.

xou816:
Hi there, thanks for the contribution :) maybe your local rust toolchain is not up to date?

src/player/player.rs
```diff
@@ -223,6 +223,8 @@ playlist-modify-public,\
 playlist-modify-private,\
 streaming";
 
+const KNOWN_AP_PORTS: &[u16] = &[80, 443, 4070];
```
xou816, line 226: You can have your Option there, and have it be owned as well since there are no heap allocations needed:

```rust
const KNOWN_AP_PORTS: [Option<u16>; 4] = [None, Some(80), Some(443), Some(4070)];
```

src/player/player.rs
```diff
@@ -240,15 +242,50 @@ async fn create_session(
...
+            for port in ports {
+                let session_config = SessionConfig {
+                    ap_port: port,
+                    ..Default::default()
+                };
+                let result = Session::connect(session_config, credentials.clone(), None).await;
+
+                match result {
+                    Ok(session) => {
+                        ret = Some(Ok(session));
```
xou816, line 271: why do you have to wrap the result in an Option? When is it a None?

(also I think you can `break Ok(session)` for instance, and that would be the value of your match block)

src/player/player.rs
```diff
@@ -240,15 +242,50 @@ async fn create_session(
...
+            result.map_err(|e| {
+                dbg!(e);
+                SpotifyError::LoginFailed
+            })
+        },
+        None => {
+            let mut ret = None;
+
+            let mut ports: Vec<Option<u16>> = KNOWN_AP_PORTS.iter().map(|port| Some(port.to_owned())).collect();
+            ports.push(None);
```
xou816, line 261: Shouldn't the default port be tried first?

xou816 (changes requested):

src/player/player.rs
sei0o: Fixed in the latest commit (`KNOWN_AP_PORTS`'s first element is now `None`). Thank you for helpful reviews :)

src/player/player.rs
sei0o: I meant `ret` is None if there is not any access point available, but I just replaced `ret` with `return` and this looks better (see the latest commit).

The compiler [told](https://doc.rust-lang.org/stable/reference/expressions/loop-expr.html#break-and-loop-values) me you can break with values in `loop {...}` and breakable blocks, but not inside `for`-loop.

src/player/player.rs
sei0o: Great. Fixed in the latest commit.

xou816:
Perfect, LGTM, merging! thanks for your work!

PR #366, opened 2021-11-10
Details Redesign 

### LINKED ISSUE #348: Use standard list styling
Lists use this weird borderless style:
image
It would be better to use standard ~~.content~~ .boxed-list style class on these.

jannuary:
Resolves #348 

Uses standard styles where applicable, gives songs more space, uses multipliers of 6 for margins, makes album cover larger, uses real icon for like button, ~~adds an animated playback indicator~~ (#358), ~~makes the entire window scrollable~~ and tries to replicate the boxed-list style for songlist. 
image
image
image

TODO:

- ~~[ ] Pause playback indicator with the song~~ will be in a followup
- [x] Make album header adaptive
- - [x] Ditto for playlist 
- [x] Kill custom styles and classes where possible 
- [x] Use boxed-list-ish style on queue and saved songs
- - [ ] ~~Add initial state for queue~~ will be in a followup
- [x] Ditto for artist view lists
- [x] Add border and rounded corners to album cover
- [ ] ~~Add inline release details?~~


xou816:
I'm so glad someone with a good eye for design is contributing honestly :D thanks!

before you head too far down that road: if you can, keep that hide-header-on-scroll thing, it's needed so that the Gtk ListView can manage long lists (AFAIK it has to be the only child of a scrolled window for that to work, I'll try to find a reference to back this up)

jannuary:
Hm, if true, that really sucks :/ 
I've noticed in my testing that the list falls apart on 200th item, but interacting with it seems to bring it back

jannuary:
Alright, reverted that for now. 
I suppose if we have to have a scrolling listview we could have a header like this, though that requires a split headerbar we don't have yet:
dsdtdta 1

xou816:
Wow this looks super good! Love the idea! I really need to get to work :D I think I want to rework selection first, but then I might take a stab at the header! For now if that's okay with you lets just have the header work as it does now

xou816:
> Seems like a relevant issue: [listview scrolling broken](https://gitlab.gnome.org/GNOME/gtk/-/issues/2971#note_1250397)

I did encounter this one, but it's somewhat unrelated -- regardless of that issue, I am pretty sure the listview needs to be the only child of the scrollview.

jacksongoode:
> Alright, reverted that for now. I suppose if we have to have a scrolling listview we could have a header like this, though that requires a split headerbar we don't have yet: dsdtdta 1

This looks really amazing! The dynamic titlebar is a nice touch as well to minimize the content in the window on scroll.

jannuary:
For now, I deleted the header from saved tracks/now playing screens. This shouldn't be too much of a regression given sidebar clearly shows the selected screen, but eventually we should have a title in the headerbar.

xou816:
Hmm okay i'll work on the new headers very soon anyway but why did you need to do this out of curiosity? 😊

jannuary:
It's kinda visually awkward to have this overlaid header with a large thin label. Smaller bold header on top of the list could've worked a bit better but since that's not very feasible with listviews, so I feel it makes sense to just remove it and delegate to the headerbar.

jannuary:
Started work on this: 
Kooha-11-13-2021-21-04-23


Unfortunately, can't easily reuse widgets here, so I suppose I should create some sort of template for the header, help for which would be greatly appreciated.

xou816:
Off to a good start!! :+1: 

For the actual header, I'll probably implement a simple split headerbar in my branch, then we can actually to implement your fancy mockup above! I should have something later today :)

As for your adaptative header, I think you're on the right track! I'd recommend following a similar pattern to this one: [some wrapper widget](https://github.com/xou816/spot/blob/development/src/app/components/playback/playback_widget.rs#L25) could have both the regular and mobile version in an adwaita squeezer, and then this wrapper widget [manages both versions](https://github.com/xou816/spot/blob/development/src/app/components/playback/playback_widget.rs#L90). This way you're free to have 2 versions of the widget, potentially very different, while still making sure both are synced. I don't know if this answers your question?

src/app/components/playlist/song.css
```diff
@@ -58,7 +58,6 @@
 
 row {
   transition: background-color 150ms ease;
```
jannuary, line 60: oh that also fixes weird sidebar spacing

two birds with one stone i guess

jannuary:
I have no idea why this doesn't work, but now the album header is a template separate from the details screen.

xou816:
Checkout out your branch to see what was going on: I love it, looks so much better! Pretty smart to play with orientation property too actually :) 

As to why it doesn't work: you're creating a new instance of your AlbumHeaderWidget instead of using the one that's created by the AlbumDetailsWidget -- grab it from here instead of creating+storing it and it works!

jannuary:
Sorry, but could you point out in code how to specifically use the created AlbumHeaderWidget? I'm trying to do so, but it doesn't build that way.

xou816:
Oh, sure! I would recommend adding the following method to the AlbumDetailsWidget:
```rust
fn header_widget(&self) -> &AlbumHeaderWidget {
    &self.widget().header_widget
}
```

jannuary:
This works, thanks a lot!
Kooha-11-15-2021-09-23-05

xou816:
played with both our branches, must say I really like where this is headed!!

image

maybe the art is too large though, especially on mobile devices (takes more than half of the screen)

src/app/components/details/details.css
```diff
@@ -1,44 +0,0 @@
-label.details__album-label {
```
xou816, line 1: I prefer that we keep these local css files to make it easier to maintain styles if you don't mind

jacksongoode:
> This works, thanks a lot! Kooha-11-15-2021-09-23-05

Should there be gutter padding on the sides of the track list at mobile widths? It might make sense to drop the rounded corners or extend the tracks to the edges of the window. What do you think @jannuary?

jannuary:
> Should there be gutter padding on the sides of the track list at mobile widths? It might make sense to drop the rounded corners or extend the tracks to the edges of the window. What do you think @jannuary?

It's not a very common pattern in GNOME apps, and would involve quite a bit of custom stuff.

src/app/components/details/details.css
jannuary: Some of these styles will just be deleted and some will be moved to app.css, like the song list styling, because it's used across multiple components.

jannuary:
I'm gonna downscale this MR because it's already quite big and finish it as a follow-up.

src/app/components/details/details.css
xou816: but the playlist details thingy is definitely specific to this component :/

src/app/components/details/details.rs
```diff
@@ -193,26 +150,47 @@ impl Details {
 
         let modal = ReleaseDetailsWindow::new();
 
-        widget.connect_liked(clone!(@weak model => move || model.toggle_save_album()));
+        widget
+            .header_widget()
+            .connect_liked(clone!(@weak model => move || model.toggle_save_album()));
```
xou816, line 155: might be worth adding a connect_liked fn to the details widget

src/app/components/details/details.rs
```diff
@@ -193,26 +150,47 @@ impl Details {
...
-                    .root()
-                    .and_then(|r| r.downcast::<gtk::Window>().ok())
-                    .as_ref(),
-            );
-            modal.show();
-        }));
-
+        widget
+            .header_widget()
+            .connect_info(clone!(@weak modal, @weak widget => move || {
```
xou816, line 169: might be worth adding a connect_info fn to the details widget

src/app/components/details/details.rs
```diff
@@ -234,13 +213,26 @@ impl Details {
...
+
             self.widget
+                .header_widget()
                 .set_album_and_artist(&album.title[..], &album.artists_name());
             self.widget
-                .connect_artist_clicked(clone!(@weak self.model as model => move || {
+                .header_mobile()
+                .set_album_and_artist(&album.title[..], &album.artists_name());
+
+            self.widget.header_widget().connect_artist_clicked(
```
xou816, line 226: might be worth adding a connect_artist_clicked fn to the details widget

src/app/components/details/details.rs
```diff
@@ -254,16 +246,29 @@ impl Details {
...
+                        Some(header_mobile),
+                        Some(ref pixbuf),
+                    ) = (
+                        widget.upgrade(),
+                        modal.upgrade(),
+                        header.upgrade(),
+                        header_mobile.upgrade(),
+                        pixbuf,
+                    ) {
+                        header.set_artwork(pixbuf);
```
xou816, line 270: might be worth adding a set_artwork fn to the details widget, so that you dont  have to upgrade two headers here as well

src/app/components/details/details.ui
```diff
@@ -8,113 +8,30 @@
...
-                        <property name="valign">center</property>
-                        <style>
-                          <class name="circular"/>
-                          <class name="details__like"/>
-                        </style>
-                      </object>
-                    </child>
-                  </object>
+                <object class="AdwSqueezer">
+								<property name="valign">center</property>
```
xou816, line 22: (if you dont mind reformating the xml that'd be nice!)

src/app/components/playlist/song.ui
```diff
@@ -111,13 +110,14 @@
...
         <property name="halign">end</property>
+				<property name="valign">center</property>
         <layout>
           <property name="row-span">2</property>
           <property name="column">4</property>
           <property name="row">0</property>
         </layout>
         <style>
-          <class name="song__menu"/>
+          <class name="circular"/>
```
xou816, line 120: thanks! better with standard classes indeed :)

src/app/components/playlist_details/playlist_details.rs
```diff
@@ -184,9 +148,21 @@ impl PlaylistDetails {
             model.load_more_tracks();
         }));
 
-        widget.connect_owner_clicked(clone!(@weak model => move || {
-            model.view_owner();
-        }));
+        widget
+            .header_widget()
+            .connect_artist_clicked(clone!(@weak model => move || {
```
xou816, line 153: similar comments to the above

src/spot.gresource.xml
```diff
@@ -13,11 +13,11 @@
     <!-- artist -->
     <file alias="components/artist.ui">app/components/artist/artist.ui</file>
     <!-- details -->
-    <file alias="components/details.css">app/components/details/details.css</file>
     <file alias="components/details.ui">app/components/details/details.ui</file>
+		<file alias="components/album_header.ui">app/components/details/album_header.ui</file>
+		<file alias="components/album_header.css">app/components/details/album_header.css</file>
```
xou816, line 18: file does not exist apparently, forgot to commit it? removed it in the meantime maybe?

xou816:
this looks really good :) maybe the cover arts are a bit large though? especially on mobile, you can only see one or two songs at most

xou816:
i think there's a small issue with the clamp and short titles as well: 
image

xou816:
oh and looks like something is off with the alignment or is that just me? 
image

src/spot.gresource.xml
```diff
@@ -50,25 +51,5 @@
   <gresource prefix="/dev/alextren/Spot/icons/scalable/actions/">
     <file compressed="true" preprocess="xml-stripblanks" alias="music-queue-symbolic.svg">app/components/selection/icons/music-queue-symbolic.svg</file>
     <file compressed="true" preprocess="xml-stripblanks" alias="playlist2-symbolic.svg">app/components/selection/icons/playlist2-symbolic.svg</file>
-  </gresource>
-	<gresource prefix="/dev/alextren/Spot/icons/scalable/actions/">
```
xou816: oh :( are we not doing this for now? maybe later?

src/spot.gresource.xml
jannuary: Apparently :\
This also fixes the issues you commented on the PR itself.

src/spot.gresource.xml
jannuary: Yeah, this will be shipped as a separate MR.

jannuary:
> this looks really good :) maybe the cover arts are a bit large though? especially on mobile, you can only see one or two songs at most

I set it to 160, hope that's fine.

> i think there's a small issue with the clamp and short titles as well: image

I'm not too sure what's a clean way to solve it - Squeezer only switches its children when there's not enough space. In this case, since "?" can always fit, it simply never switches.
In the meantime, for playlists, I set the switch policy to "natural", so this should happen less often.

jannuary:
> but the playlist details thingy is definitely specific to this component :/

Github doesn't let me comment here, but `label.details__album-label` has been replaced with title-1, so this is no longer necessary.

src/app.css
```diff
@@ -16,4 +16,19 @@ headerbar.selection-mode {
 
 button.tool, menubutton.tool {
```
jannuary, line 17: We don't need these anymore.

src/app/components/details/release_details.rs
```diff
@@ -65,7 +65,7 @@ mod imp {
 }
 
 glib::wrapper! {
-    pub struct ReleaseDetailsWindow(ObjectSubclass<imp::ReleaseDetailsWindow>) @extends gtk::Widget, libadwaita::Window;
+       pub struct ReleaseDetailsWindow(ObjectSubclass<imp::ReleaseDetailsWindow>) @extends gtk::Widget, libadwaita::Window;
```
jannuary, line 68: What happened here?

src/app/components/details/album_header.rs
```diff
@@ -0,0 +1,127 @@
...
+        let widget = self.widget();
+        widget.album_label.set_label(album);
+        widget.artist_button_label.set_label(artist);
+    }
+
+    pub fn set_centered(&self) {
+        let widget = self.widget();
+        widget.album_label.set_halign(gtk::Align::Center);
+        widget.artist_button.set_halign(gtk::Align::Center);
+        widget.like_button.set_margin_end(0);
```
jannuary, line 118: This could honestly be in css

src/app/components/details/album_header.ui
```diff
@@ -0,0 +1,101 @@
...
+            </object>
+            </child>
+            <child type="overlay">
+            <object class="GtkButton" id="info_button">
+                <property name="icon-name">preferences-system-details-symbolic</property>
+                <property name="halign">end</property>
+                <property name="valign">end</property>
+                <property name="margin-start">6</property>
+                <property name="margin-end">6</property>
+                <property name="margin-top">6</property>
```
jannuary, line 36: Do we need this?

xou816:
> I'm not too sure what's a clean way to solve it - Squeezer only switches its children when there's not enough space. In this case, since "?" can always fit, it simply never switches.
In the meantime, for playlists, I set the switch policy to "natural", so this should happen less often.

fine by me! :) there's probably something off with the link button, its ellipsizing by default but this is the kind of things we can figure out later

src/app/components/details/release_details.rs
xou816: idk, feel free to revert !

xou816: (unless rustfmt forces you to keep it that is)

src/app/components/details/album_header.rs
xou816: or the ui file yeah

src/app/components/details/album_header.ui
xou816: I think so? so that the info button does not sit right in the corner of the image, but has some margin

src/app.css
xou816: (feel free to remove, if you don't I'll do that on some other branch eventually)

src/app/components/details/details.rs
xou816: (I can definitely help you with these not so fun tasks if needed)

src/app/components/details/album_header.ui
jannuary: Ah, nevermind, yes.

src/app/components/artist_details/artist_details.ui
```diff
@@ -32,8 +32,11 @@
                 <child>
                   <object class="GtkListView" id="top_tracks">
                     <style>
-                      <class name="artist__top-tracks"/>
+                      <class name="song__list"/>
```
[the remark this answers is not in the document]
jannuary: Turns I don't even need this style class all that much - I removed that and applied the style to .playlist instead.

src/app/components/details/details.rs
jannuary: Yeah, that would be great if you don't mind.

xou816:
Hey! :) need any help with that? If you're fine with me pushing to your branch I could help with the details I mentioned in the review!

jannuary:
> Hey! :) need any help with that? If you're fine with me pushing to your branch I could help with the details I mentioned in the review!

Yeah, that would be welcome, thanks!

jannuary:
One last minute change:
image

Release details dialog now uses AdwPreferencesWindow and ActionRows.

jannuary:
Couple inspector hacks later:
image

This is gonna have to wait till this PR is merged though.

xou816:
On it! let's merge this asap :)

xou816:
> One last minute change: image
> 
> Release details dialog now uses AdwPreferencesWindow and ActionRows.

I'm not sure how I feel about this -- why Preferences? is there no way to use this kind of widget without using a Preferences window? It's glitching weirdly for me as well, an icon will show up briefly in the headerbar

jannuary:
> I'm not sure how I feel about this -- why Preferences?

It's not that it's necessary to use PreferencesWindow specifically - it just provides a simple way to create a dialog like this with boxed lists and its headers. I think Fragments v2 is using something similar for its Statistics view.

xou816:
Hmm okay sorry but I'll change that, and we can discuss it later if you want, but let's stick to the original goal of this PR first :)

xou816:
I'll let you have a look but I think we're okay to go for now? :)

jannuary:
Well, if you insist :p 
The release details dialog now doesn't close on Esc, but I think it's fine to merge now.

src/app/components/details/release_details.ui
```diff
@@ -1,129 +1,83 @@
...
+          <object class="GtkListBox">
+            <property name="margin-start">6</property>
+            <property name="margin-end">6</property>
+            <property name="margin-top">6</property>
+            <property name="margin-bottom">6</property>
+            <property name="valign">start</property>
+            <property name="selection-mode">none</property>
+            <property name="show-separators">1</property>
+            <style>
+              <class name="card" />
```
jannuary, line 33: I'm late, but you use .card for everything *but* lists - boxed-list would be better here.

Issue #354, opened 2021-10-29
Improve Selection Mode


jannuary:
Selection mode has several issues:

- It is available everywhere, not just where it applies
- The widgets in the headerbar are kept as is, which is unexpected and leads to the user being able to leave the current screen without also cancelling selection mode
- The icons are off brand and confusing, it's not intuitive what ☑, - or + does here:
![image](https://user-images.githubusercontent.com/27908024/139424872-443199e6-c680-4420-a6fa-78a8ac94d5f1.png)
- The rows in queue aren't draggable, instead opting for awkward ˅/˄ buttons; playlists aren't reordeable at all (though i suppose it's a different issue)

Here's a mockup of my redesign proposal for the selection mode, based off Allan Day's Photos redesign:
![Selmode](https://user-images.githubusercontent.com/27908024/140286338-b0cb0f87-72ed-4aca-9d9e-5c26d6a3e40a.png)

xou816:
Some of it is intentional -- like being able to search while a selection is active, so that you don't have to reenter selection mode again. Maybe there's a way this could be clearer! The selection mode being canceled is from a poor implementation of navigation, there's some work to do here as well. Ah well!

Maybe I should just drop this idea of being able to keep a selection active while browsing, but this is something that's always bothered me in most music players when building a new playlist for instance.

--

Something I'm wondering about: I really like the look of the action bar there, do you think it should cover the controls though?

jannuary:
It does seem pretty niche to have selection mode on while browsing, so I don't think it's really worth it. 

> Something I'm wondering about: I really like the look of the action bar there, do you think it should cover the controls though?

I think it makes sense to not show playback controls while selecting - that's what Music does, for example.

jannuary:
> I think it makes sense to not show playback controls while selecting - that's what Music does, for example.

Spotify too:
![Screenshot_20211031-230817_Spotify.png](https://user-images.githubusercontent.com/27908024/139592390-e88d912c-cefb-4a19-915c-1e7674efb330.png)

xou816:
The mockups are really gorgeous, I'll try to get started on the selection tools part soon enough!

xou816:
![image](https://user-images.githubusercontent.com/6277494/141382951-b6b5051f-c805-40c2-a118-7057b01293b0.png)

started working on this! I might actually work on the headerbars as I tackle this since I'll need to change them anyway :)

xou816:
[...]

PR #370, opened 2021-11-14
354 improve selection mode

xou816:
first shot at #354 ! more work needed in the future but that's a start :)

what's missing
- favoriting songs (as this was not previously implemented)
- specific editable playlist behavior (as this was not implemented)
- drag n drop


src/app/components/headerbar/headerbar.ui
```diff
@@ -0,0 +1,65 @@
...
+            <child type="title">
+              <object class="AdwWindowTitle" id="title">
+                <property name="visible">0</property>
+                <property name="title" translatable="no">Spot</property>
+              </object>
+            </child>
+            <child type="end">
+              <object class="GtkButton" id="start_selection">
+                <property name="receives-default">1</property>
+                <property name="label">Select...</property>
```
jannuary, line 30: Any reason this is a text button instead of an icon with `selection-mode-symbolic`?

src/app/components/selection/selection_toolbar.ui
```diff
@@ -0,0 +1,87 @@
...
+            </child>
+            <child>
+              <object class="GtkButton" id="move_down">
+                <property name="icon-name">go-down-symbolic</property>
+              </object>
+            </child>
+          </object>
+        </child>
+        <child type="end">
+          <object class="GtkButton" id="queue">
```
jannuary, line 31: Could [AdwButtonContent](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ButtonContent.html) be used here?

src/app/components/selection/selection_toolbar.ui
```diff
@@ -0,0 +1,87 @@
...
+          <object class="GtkMenuButton" id="add">
+            <property name="valign">center</property>
+            <property name="has-frame">0</property>
+            <!-- <property name="icon-name">playlist2-symbolic</property> -->
+            <property name="label">Add to playlist...</property>
+            <property name="direction">up</property>
+          </object>
+        </child>
+        <child type="end">
+          <object class="GtkButton" id="remove">
```
jannuary, line 61: Ditto

src/app/components/selection/selection_toolbar.ui
```diff
@@ -0,0 +1,87 @@
...
+            <property name="valign">center</property>
+            <property name="has-frame">0</property>
+            <child>
+              <object class="GtkBox">
```
jannuary, line 65: Ditto

src/app/components/selection/selection_toolbar.ui
```diff
@@ -0,0 +1,87 @@
...
+                <property name="spacing">8</property>
+                <child>
+                  <object class="GtkImage">
+                    <property name="icon-name">app-remove-symbolic</property>
```
jannuary, line 69: I wonder why this doesn't show up on my end - Icon Library lists this as preinstalled.
Either way, `user-trash-symbolic` is probably safer to use.

src/app/components/headerbar/headerbar.ui
```diff
@@ -0,0 +1,65 @@
...
+              </object>
+            </child>
+          </object>
+        </child>
+        <child type="overlay">
+          <object class="AdwHeaderBar" id="selection_header">
+            <property name="show-end-title-buttons">1</property>
```
jannuary, line 37: This should be off.

src/app/components/headerbar/headerbar.ui
xou816: Thought the intent was clearer with a label, but a full button on every page kinda feels overwhelming so you might right about going with the icon only. what do you think, is the intent of the icon clear enough?

src/app/components/selection/selection_toolbar.ui
xou816: ooooh neat, did not know this was a thing. I guess I might have to update the bindings version, will have a look thanks!

src/app/components/selection/selection_toolbar.ui
xou816: well trash is for the actual trash, not for an action, so idk. maybe there's yet another one we could use if this one has been removed for some reason?

src/app/components/headerbar/headerbar.ui
xou816: we need controls don´t we?!

src/app/components/selection/selection_toolbar.ui
xou816: edit-delete-symbolic maybe?

jannuary: Hmm, maybe `edit-delete-symbolic` for queues and `user-trash-symbolic` for playlist? FWIW Software, for example, uses trash for "uninstall".

src/app/components/headerbar/headerbar.ui
jannuary: It's expected that within the context of selection mode, window controls aren't shown.

src/app/components/headerbar/headerbar.ui
jannuary: I think the checkmark makes sense.

xou816: Alright! I'll just keep the one that's on stable gtk for now, seeing the unavailable icon while working is itching me :p

src/app/components/selection/selection_toolbar.ui
xou816: Fair enough I might change!

src/app/components/headerbar/headerbar.ui
xou816: Oh alright I missed the context! That seems reasonable

src/app/components/selection/selection_toolbar.ui
jannuary: Slightly related, it'd be nice to have a [toast](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Toast.html) allowing the user to undo deleting a song, though it isn't a blocker.

xou816:
If that's okay I'll merge this for now, and I'll fix the remaining issues with search on mobile in a separate PR

src/app/components/headerbar/headerbar.css
```diff
@@ -0,0 +1,5 @@
+.headerbar--translucent, .headerbar--translucent windowhandle {
```
jannuary, line 1: For future reference: .flat already does exactly what you wanted to do here.

xou816: oh okay did not know that, I'll fix it! do you have a reference of css classes you could share?

jannuary: Libadwaita demo has a section on style classes.

PR #383, opened 2021-11-24
Use tabular numbers, ratio symbol for durations

jannuary:
Very subtle change to how song durations are displayed: U+2236 Ratio is used instead of colon, .numeric style is applied to make the numbers tabular (as in having the numbers take the same space, allowing them to align better), as per HIG.

image

### DIFF
--- a/src/app/components/playback/playback_widget.rs
+++ b/src/app/components/playback/playback_widget.rs
@@ -112,7 +112,7 @@ impl PlaybackWidget {
             style_context.add_class(class);
             widget.seek_bar.set_range(0.0, duration);
             widget.seek_bar.set_value(0.0);
-            widget.track_position.set_text("0:00");
+            widget.track_position.set_text("0∶00");
             widget
                 .track_duration
                 .set_text(&format!(" / {}", format_duration(duration)));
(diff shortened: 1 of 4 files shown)


src/app/components/playback/playback_widget.rs
```diff
@@ -112,7 +112,7 @@ impl PlaybackWidget {
             style_context.add_class(class);
             widget.seek_bar.set_range(0.0, duration);
             widget.seek_bar.set_value(0.0);
-            widget.track_position.set_text("0:00");
+            widget.track_position.set_text("0∶00");
```
shlyakpavel, line 115: I'm not sure it's okay to put non-ansi symbols inside the code, especially if they can confuse people being much like other symbols. Maybe it should be better defined by symbol id or somewhere inside translations?

xou816: Fair point! I think it's fine for now, and it might be a bit tricky since it appears both in source code and in .ui files. I'll accept the PR as is but we could clean this up eventually (it should be possible to not have these kinds of timestamps in the .ui files at all).

PR #391, opened 2021-12-02
Add album year to the list & the details view

### LINKED ISSUE #137: Add album year to the album list
Hello there, I'm trying both your app and the ncspot these days and I just noticed that both dont show the year (or give us a way to sort the albums by release date). I think it'ld help a lot discovering music through the app. Or maybe create an option, if people feel it'll clutter the list (I doubt)

Cheers

sei0o:
Fixes #137 

It looks like this:
image

and at the details view:
image

The GTK part wasn't that hard, but the problem is that not only albums but also playlists makes use of `Album` through `TrackItem`. The album widget requires `AlbumModel` , `AlbumModel` requires `AlbumDescription`, and `AlbumDescription` requires `Album`. 

I tried moving `release_date` in `AlbumReleaseDetails` to `AlbumDescription` (in commit `dc71a29`) and it worked fine for albums. However, since playlists do not have their release dates,  the program fails to parse them. I could just let `release_date` stay in `AlbumReleaseDetails` and use it, but in `SearchResults::update_results`, for example, we do not have access to an  `AlbumReleaseDetails`.

The commits are messed up... I will squash them later.


src/app/models/main.rs
```diff
@@ -7,7 +8,13 @@ use crate::app::components::utils::format_duration;
 
 impl From<&AlbumDescription> for AlbumModel {
     fn from(album: &AlbumDescription) -> Self {
-        AlbumModel::new(&album.artists_name(), &album.title, &album.art, &album.id)
+        AlbumModel::new(
+            &album.artists_name(),
+            &album.title,
+            album.year(),
```
sei0o, line 14: This doesn't compile because `AlbumDescription` does not have `year()` or `released_date`. Using `FullAlbum, AlbumFullDescription` would do?

xou816:
Hey, thanks a lot for your work! :) this looks promising!

> but also playlists makes use of Album through TrackItem.

Yeah, sorry about that :( I'm absolutely open to this being changed though! Two possibilities off the top of my head: either we keep going with this model being shared but adapt the widget so that things such as the release date are `Option`s, or break it into two distinct widget+model pairs. What do you think?

As for search results or playlists, well if we don't have the date, don't sweat it, we won't be able to show it and it is fine. This too could be solved by having the release date be an Option in that AlbumModel I think!

src/app/components/details/details.rs
```diff
@@ -160,10 +163,11 @@ impl AlbumDetailsWidget {
         self.widget().album_art.set_from_pixbuf(Some(art));
     }
 
-    fn set_album_and_artist(&self, album: &str, artist: &str) {
+    fn set_album_and_artist_and_year(&self, album: &str, artist: &str, year: u32) {
```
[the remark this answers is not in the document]
xou816: It's a bit of a mouthful but that's fine, at least it's descriptive :D  Feel free to split it into multiple calls if you're more comfortable with that!

src/app/models/main.rs
xou816: We could make it an Option I suppose! (see main answer to the issue)

src/app/components/details/details.rs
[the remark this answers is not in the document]
sei0o: Then I'll leave it in this way :)

sei0o:
Thanks for the advice! Now I have wrapped `release_date` in some structs with `Option`. I was also thinking about the same idea as your second option, but for now this solution would be enough.

src/app/components/album/album.css
```diff
@@ -35,11 +35,15 @@ leaflet.folded .album .album__cover {
   font-size: 14px;
 }
 
+.album label.album__year {
+  font-size: 14px;
```
jannuary, line 39: We really shouldn't hardcode font-sizes like that. Not a blocker for this particular PR since it's used elsewhere, but eventually, we need to remove these.

src/app/components/album/album.ui
```diff
@@ -51,6 +51,18 @@
         </style>
       </object>
     </child>
+    <child>
+      <object class="GtkLabel" id="year_label">
+        <property name="label">Year</property>
+        <property name="justify">center</property>
+        <property name="wrap">1</property>
+        <property name="wrap-mode">word-char</property>
+        <property name="max-width-chars">1</property>
```
jannuary, line 60:
```suggestion
        <property name="max-width-chars">1</property>
        <property name="sensitive">0</property>
```

This will grey the year out, helping with the UI hierarchy.

src/app/components/search/search.rs
```diff
@@ -190,6 +190,7 @@ impl SearchResults {
                 self.album_results_model.append(&AlbumModel::new(
                     &album.artists_name(),
                     &album.title,
+                    album.year(),
```
jannuary, line 193: My Rust is rusty, but isn't there supposed to be an ampersand here? I think this moves album.year as opposed to referencing it.

src/app/components/album/album.ui
jannuary: Also, github doesn't let me comment there, but the `spacing` of AlbumWidget seems a bit too spacious, 6 looks a bit better to me

src/app/components/album/album.css
sei0o: Agreed. GTK offers `@define-color` for reduce hard-coded colors, but it doesn't for other types. Neither `var()` seems to be unavailable for GTK so without preprocessors would that be difficult?

jannuary: rem seems to do the trick:
image

src/app/components/search/search.rs
sei0o: `album.year()` returns an `u32` and yes, it's moved into the `AlbumModel`. `u32` is always small while strings can be sometimes larger so they are passed over references I suppose. 

`SongModel::new` also takes `u32` and its value is given without an ampersand. In this case `track_number` is implicitly copied into `SongModel` since `u32` implements `Copy`.
https://github.com/xou816/spot/blob/3fa78e951881de05a19940fe6a916a3b51a0ceda/src/app/models/main.rs#L41

src/app/components/album/album.ui
sei0o: Thanks for reviewing! It works:

image

With spacing = 2, just to make sure things are working (my eyeballs couldn't confidently tell two pixels):
image

What about the details view?
image

jannuary: Details gonna recieve a facelift soon enough anyway, so I think just making it insensitive for now is fine.

src/app/models/gtypes/album_model.rs
```diff
@@ -85,6 +98,7 @@ mod imp {
             Self {
                 album: RefCell::new(None),
                 artist: RefCell::new(None),
+                year: RefCell::new(0),
```
xou816, line 101: I feel like this could definitely be a RefCell<Option<u32>>, this would avoid all the unwrapping and awkward comparison to 0 you have to deal with!

xou816 (changes requested):
looks good to me for the most part, again thanks!! :)

src/app/models/gtypes/album_model.rs
sei0o: I was still confused how GTK deals with optional values. Fixed in the latest commit. :)

src/app/models/main.rs
```diff
@@ -99,6 +109,12 @@ impl AlbumDescription {
         let duration: u32 = self.songs.iter().map(|song| song.duration).sum();
         format_duration(duration.into())
     }
+    pub fn year(&self) -> Option<u32> {
+        match self.release_date {
+            Some(ref date) => Some(u32::from_str(date.split('-').next().unwrap()).unwrap()),
```
xou816, line 114: it is a bit unfortunate to unwrap and wrap it back up!

you could probably handle this a little bit better using [map ](https://doc.rust-lang.org/std/option/enum.Option.html#method.map)or [and_then](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then), and [ok](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok)

src/app/models/main.rs
```diff
@@ -22,6 +29,8 @@ impl From<&PlaylistDescription> for AlbumModel {
         AlbumModel::new(
             &playlist.owner.display_name,
             &playlist.title,
+            // Playlists do not have their released date since they are expected to be updated anytime.
```
xou816, line 32: 👍 helpful comment!

src/app/models/gtypes/album_model.rs
```diff
@@ -141,6 +168,7 @@ mod imp {
             match pspec.name() {
                 "album" => self.album.borrow().to_value(),
                 "artist" => self.artist.borrow().to_value(),
+                "year" => self.year.borrow().unwrap_or(0).to_value(),
```
xou816, line 171: I don't know how easy it would be because of the borrow call (always a bit tricky sadly) but unwrapping should ideally not be needed.

you might have to toy around with [Ref::map](https://doc.rust-lang.org/std/cell/struct.Ref.html#method.map)

src/app/models/gtypes/album_model.rs
```diff
@@ -45,9 +60,10 @@ mod imp {
 
     // Static array for defining the properties of the new type.
     lazy_static! {
-        static ref PROPERTIES: [glib::ParamSpec; 4] = [
+        static ref PROPERTIES: [glib::ParamSpec; 5] = [
             glib::ParamSpec::new_string("artist", "Artist", "", None, glib::ParamFlags::READWRITE),
             glib::ParamSpec::new_string("album", "Album", "", None, glib::ParamFlags::READWRITE),
+            glib::ParamSpec::new_uint("year", "Year", "", 0, 9999, 0, glib::ParamFlags::READWRITE),
```
xou816, line 66: ...hmm, okay I think I see why this is trickier than expected 😅 new_uint does not play nice with Options by the looks of it. if you want to keep that file as is I'm OK with that! the unwrap_or(0) below will do

src/app/components/album/album.rs
```diff
@@ -101,6 +104,18 @@ impl AlbumWidget {
...
+
+        match album_model.year() {
+            Some(_) => {
+                album_model
+                    .bind_property("year", &*widget.year_label, "label")
+                    .flags(glib::BindingFlags::DEFAULT | glib::BindingFlags::SYNC_CREATE)
+                    .build();
+            }
+            None => {
+                widget.year_label.hide();
```
xou816, line 116: good call! 👍

xou816 (commented):
almost there!

src/app/models/gtypes/album_model.rs
sei0o: Yeah, [numbers cannot be optional](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html), so I think it has to be unwrapped to something... I'm leaving it as is.

src/app/models/main.rs
```diff
@@ -99,6 +109,12 @@ impl AlbumDescription {
         let duration: u32 = self.songs.iter().map(|song| song.duration).sum();
         format_duration(duration.into())
     }
+    pub fn year(&self) -> Option<u32> {
+        self.release_date
+            .as_ref()
+            .and_then(|date| date.split('-').next())
+            .and_then(|y| u32::from_str(y).ok())
+    }
```
sei0o, line 117: So it looks better now

src/app/models/main.rs
sei0o: Ah, yes, that is exactly I had to use those functions. In the latest commit I'm using `and_then` and `ok` :)

src/app/models/main.rs
xou816: oh it was not so much about human friendliness, it's just that calls to unwrap() can panic!, and it's easy enough to avoid them here so...

but yeah that's better!!

src/app/models/gtypes/album_model.rs
xou816: right, fine by me :)

xou816 (approved):

xou816:
All good to me :) looks like you've got a few conflicts to fix but then we're good to merge!

sei0o:
Resolved conflicts :)

xou816:
Perfect, great work, merging this. thanks!

PR #393, opened 2021-12-04
Define accent_color as well

jannuary:
Apparently needed for things like focus rings:
image


xou816:
Trusting you on that, thanks for the fix! :)

xou816:
Oops, I was a bit fast, there's a missing semi colon and rightfully so a warning happening

> (process:36443): Gtk-WARNING **: 17:48:02.193: Theme parsing error: app.css:3:1-2: Missing semicolon at end of color definition

PR #395, opened 2021-12-07
Use relative font-size

jannuary:
Instead of setting font-size in pixels, define it relative to rem. (by default 14px)

Makes text follow user's font scale preference:
image


xou816:
and I pretend to be a webdev :facepalm: thanks, as always! merging!

PR #399, opened 2021-12-10
update album view

jannuary:
- Use .card for album cover
- Remove custom styles
- Rework hover 
- Make entire thing clickable
image

### DIFF
--- a/src/app/components/album/album.css
+++ b/src/app/components/album/album.css
@@ -3,6 +3,7 @@
 leaflet.unfolded .album .card {
   min-width: 200px;
   min-height: 200px;
+  border-radius: 6px;
 }
 
 leaflet.unfolded .album {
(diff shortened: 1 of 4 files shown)


xou816:
I quite liked the old hover effect, maybe we could have both? :blush: 

Nice work otherwise, makes sense to make the whole thing clickable.

spacing_card

The spacing around the image seems a bit inconsistent to me on hover, and maybe the corner radii should match?

xou816:
oh right :( my bad! 

we still can probably do something about it, if we add a container around the button, so that the container grows but the button is centered within it

jannuary:
Should be better now.

xou816:
oh it is much better indeed! LGTM, merging if you're good to go on your end?

jannuary:
Yeah, should be fine to merge.

PR #402, opened 2021-12-13
Update lists, visually merge album header

### LINKED ISSUE #379: Use AdwClampScrollable
Just a reminder to myself to make ListView views use [AdwClampScrollable](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ClampScrollable.html) instead of AdwClamp.

jannuary:
- Don't use .card, use custom style for playlists instead
One thing about the previous MR is that .card applies to the entire ListView, including stuff like rounded corners. Herein, I try to recreate the .boxed-list style, applying the style to individual rows.
- Use AdwClampScrollable
In practice, this means that the scrollbar is where it's supposed to be while technically being under ListView ∈ ScrolledWindow rule.
- Visually merge album header & headerbar
Make header revealing/collapsing feel more intentional. 

image

Resolves #379


xou816:
I think I preferred the color scheme in your original mockup (darker header, lighter list)

TIL about AdwScrollableClamp, nice! is the ScrollableWindow needed at all in that case?

jannuary:
> I think I preferred the color scheme in your original mockup (darker header, lighter list)

Does view_bg look fine? 
image
Thing is, we can't exactly have a 'darker' color without having a totally separate light and dark css.

> TIL about AdwScrollableClamp, nice! is the ScrollableWindow needed at all in that case?

Clamp doesn't implement any of the scrolling business afaik, it's just a helper widget of sorts.

jannuary:
> Thing is, we can't exactly have a 'darker' color without having a totally separate light and dark css.

I guess we could use AdwApplication - it provides separate css for free, more or less.

jannuary:
> 
> > Thing is, we can't exactly have a 'darker' color without having a totally separate light and dark css.
> 
> I guess we could use AdwApplication - it provides separate css for free, more or less. 
> 
> 


Also, there has to be some sorta transition between header header and headerbar header.

xou816:
> Also, there has to be some sorta transition between header header and headerbar header.

what kind of transition are you thinking of, is it about having the title appear in the headerbar?

> I guess we could use AdwApplication - it provides separate css for free, more or less.

I'll have a look, not sure what this allows! don't worry too much about the bg color otherwise, if you like it that way I think I can live with it :D

jannuary:
One more thing - I want to make the entire header area draggable, but that conflicts with the swipe tracker - clicking on the header area immideately makes it collapse. Anything that could be done about that?

xou816:
Clicking should not make it collapse, it should not identify a simple click as a swipe so there could be a small issue here, but yes, absolutely we could make it touch only! I am not sure how that would work with a GtkWindowHandle which I assume you´d want here?

xou816:
Actually I'm saying we can make it touch only but I can't remember how at the moment... I think there was a flag for that in gtk3...

jannuary:
> Clicking should not make it collapse, it should not identify a simple click as a swipe

Yeah, pardon, it doesn't indeed - I meant it collapses on drag.

xou816:
Oh yeah okay checked out your branch (looking good!!) and I see what you mean :/

not sure how to work around that! Except to make it touch only indeed

xou816:
Well [set_touch_only](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/prelude/trait.GestureSingleExt.html#tymethod.set_touch_only`) seems to do the trick :D

xou816:
> Well [set_touch_only](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/prelude/trait.GestureSingleExt.html#tymethod.set_touch_only%60) seems to do the trick :D

(diff for reference)

```diff
--- a/src/app/components/details/details.rs
+++ b/src/app/components/details/details.rs
@@ -97,9 +97,11 @@ impl AlbumDetailsWidget {
         );
 
         let swipe_controller = gtk::GestureSwipe::new();
+        swipe_controller.set_touch_only(true);
         swipe_controller.set_propagation_phase(gtk::PropagationPhase::Capture);
         swipe_controller.connect_swipe(clone!(@weak self as _self => move |_, _, dy| {
             _self.set_header_visible(dy >= 0f64);
+            _self.set_header_visible(dy > -0.1);
         }));
 
         self.widget()
```

I see you're still making changes to the style, just give me a heads up when you want to merge things:)

jannuary:
Have unpushed commits and several other nitpicks and also not at home, so making a draft

xou816:
Perfect 👍

jannuary:
Alright, outside of the headerbar transition, I'm happy with this now.

xou816:
I have one or two things I need to do with headerbars anyway, maybe I could help! what did you have in mind again? Have the album title appear in the header when scrolling is that right?

jannuary:
Removing .flat and showing the title when the header is hidden, yeah

xou816:
https://github.com/xou816/spot/compare/listing-headerbar here is an update for that! I should not have made this Screen-wrapper thing but long story short, should be easier to use our selection-enabled headerbars now

if that looks okay to you I'll push to your branch!

image

image

not so sure how to make the css transition good but it's a start

src/app/components/details/album_header.css
```diff
@@ -1,14 +1,21 @@
...
 }
 
 .title-4 {
   padding: 0;
 }
 
+
+.title-4:hover {
+  border-radius: 6px;
+  background-image: image(alpha(currentColor, 0.03));
```
xou816, line 14: a bit too subtle maybe? I had not even noticed it before starting reading the code :D

src/app/components/details/album_header.css
```diff
@@ -1,14 +1,21 @@
 .title-4 label {
```
xou816, line 1: is title-4 a standard class? I think it is! if that's the case, I'd rather have us define an additional class for the artist link, so that we don't have affect other possible uses of title-4 in the future

xou816:
Two small things: 

image

I think we could use some spacing between the headerbar and the content

image

Maybe there should be a border there as well?

src/app/components/details/details.rs
```diff
@@ -87,7 +87,7 @@ impl AlbumDetailsWidget {
         let is_up_to_date = widget.header_revealer.reveals_child() == visible;
         if !is_up_to_date {
             widget.header_revealer.set_reveal_child(visible);
-            widget.headerbar.set_title_visible(!visible);
+            widget.headerbar.set_title_visible(true);
```
xou816, line 90: think the line can be removed altogether as well as the property in the ui file if we're using opacity now :)

jannuary: Yeah, this is a dirty hack - title was invisible if not explicitly set visible somehow.

jannuary:
> Maybe there should be a border there as well?

Not sure why it doesn't - I removed the border from the main headerbar as to not interfere with the clamp's but it shouldn't affect selection mode?

jannuary:
Yeah, I'm not touching its box-shadow at all, it just doesn't get rendered for some reason

xou816:
ah well, we'll figure this out later! thank you so much, those screens you've been working on look so much better than the others!!

think it might be time to release a new version on master too, if its compatible with the currently released runtime!

PR #406, opened 2021-12-19
Add volume control

Diegovsky:
Currently it works via the mpris interface but allows implementing via a GUI slider futurely.

I added some new events and an action as I was groking the codebase so I'm not sure if I edited the wrong places

### DIFF
--- a/src/dbus/mpris.rs
+++ b/src/dbus/mpris.rs
@@ -9,7 +9,7 @@ use zbus::ObjectServer;
 use zvariant::ObjectPath;
 
 use super::types::*;
-use crate::app::state::RepeatMode;
+use crate::app::state::{PlaybackEvent, RepeatMode};
 use crate::app::{state::PlaybackAction, AppAction};
 
 #[derive(Clone)]
@@ -298,7 +298,7 @@ impl SpotMprisPlayer {
     }
 
     #[dbus_interface(property)]
-    pub fn set_rate(&self, value: f64) {}
+    pub fn set_rate(&self, value: f64) { }
 
     #[dbus_interface(property)]
     pub fn shuffle(&self) -> bool {
@@ -314,9 +314,14 @@ impl SpotMprisPlayer {
 
     #[dbus_interface(property)]
     pub fn volume(&self) -> f64 {
-        0f64
+        self.state.volume()
     }
 
     #[dbus_interface(property)]
-    pub fn set_volume(&self, value: f64) {}
+    pub fn set_volume(&self, value: f64) -> Result<()> {
+        self.state.set_volume(value);
+        self.sender
+            .unbounded_send(PlaybackAction::SetVolume(value).into())
+            .map_err(|_| zbus::fdo::Error::Failed("Could not send action".to_string()))
+    }
 }
--- a/src/dbus/mpris.rs
+++ b/src/dbus/mpris.rs
@@ -298,7 +298,7 @@ impl SpotMprisPlayer {
     }
 
     #[dbus_interface(property)]
-    pub fn set_rate(&self, value: f64) { }
+    pub fn set_rate(&self, value: f64) {}
 
     #[dbus_interface(property)]
     pub fn shuffle(&self) -> bool {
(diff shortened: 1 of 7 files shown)


src/dbus/types.rs
```diff
@@ -167,9 +168,23 @@ impl SharedMprisState {
...
     }
 
+    pub fn volume(&self) -> f64 {
+        self.0
+            .lock()
+            .map(|s| s.volume)
+            .map_err(|e| error!("Failed to get volume: {:?}", e))
+            .unwrap_or(1.0f64)
+    }
+    pub fn set_volume(&self, volume: f64) {
```
xou816, line 182: (nitpick) could use a space above the function

src/dbus/mpris.rs
```diff
@@ -314,9 +314,14 @@ impl SpotMprisPlayer {
...
     #[dbus_interface(property)]
     pub fn volume(&self) -> f64 {
-        0f64
+        self.state.volume()
     }
 
     #[dbus_interface(property)]
-    pub fn set_volume(&self, value: f64) {}
+    pub fn set_volume(&self, value: f64) -> Result<()> {
+        self.state.set_volume(value);
```
xou816, line 322: right now its not a big deal at all, but in the long term I´d recommend not modifying the state here, but only when the change has been acknowledged by the main process (written to the app state). this keeps all mutations of the state (and dbus outbound notifications) in one place, whether this happened through mpris or through something else (like a slider) :)

src/dbus/mpris.rs
```diff
@@ -9,7 +9,7 @@ use zbus::ObjectServer;
 use zvariant::ObjectPath;
 
 use super::types::*;
-use crate::app::state::RepeatMode;
+use crate::app::state::{PlaybackEvent, RepeatMode};
```
xou816, line 12: unused import apparently

src/player/player.rs
```diff
@@ -6,10 +6,12 @@ use librespot::core::config::SessionConfig;
 use librespot::core::keymaster;
 use librespot::core::session::{Session, SessionError};
 
+use librespot::playback::mixer::softmixer::SoftMixer;
+use librespot::playback::mixer::{AudioFilter, Mixer};
```
xou816, line 10: AudioFilter is not used either apparently

xou816 (commented):
Hey! Thanks a lot for this, I had no idea how this would be implemented on the librespot side, nice! This is a really helpful feature. Looks good to me overall! Just a few minor comments

src/dbus/mpris.rs
Diegovsky: I agree! How would I get the volume from the main process from mpris?

src/dbus/mpris.rs
Diegovsky: woops must be because of my refactor

Diegovsky:
> Hey! Thanks a lot for this, I had no idea how this would be implemented on the librespot side, nice!

Yeah they don't have docs so I had to be a little creative lol

I'm honestly impressed I could pull this off :)

src/player/player.rs
Diegovsky: I'm pretty sure it's because of the trait but I'll fix it in the next commit

src/dbus/types.rs
Diegovsky: noted

src/dbus/mpris.rs
xou816: my files are really poorly organized but you can see that in the `impl EventListener for AppPlaybackStateListener`

Diegovsky: Nice! I'll be comitting my changes soon but I'll take some days to respond since it's new years

Diegovsky:
I just uploaded the new changes and also fixed a problem with librespot's SoftMixer; apparently it starts with 50% volume instead of 100% so I had to hard code it to 100% to be consistent with initial values.

xou816:
Perfect, good thinking with the clamping too!!

Thanks a lot, merging this!

Issue #287, opened 2021-08-25
[Feature request] Make better use of sidepanel, option to list playlists


jacksongoode:
Right now, unless there are plans to make more use of the side panel, I think it would be nice to have a list of playlists. This might require the introduction of some line barriers to divide the collections (Library, Playlists, and Now Playing) and the playlists themselves.

xou816:
Fair enough, that'd be handy indeed! lots of wasted space otherwise

abegert:
I have started working on this, hope that's okay. My idea is to have the playlists in an expandable treeview, which is collapsed by default.
I'm also planning on adding the already mentioned line barriers.

abegert:
> > I have started working on this, hope that's okay. My idea is to have the playlists in an expandable treeview, which is collapsed by default. I'm also planning on adding the already mentioned line barriers.
> 
> Great to hear! However, I'm not sure if having it be expandable and collapsed by default is the way to go.

What would you prefer? Moving Playlists to the bottom and then adding items for each playlist underneath?

jannuary:
> What would you prefer? Moving Playlists to the bottom and then adding items for each playlist underneath?

Yeah, that sounds good.
In more general thoughts about sidebar, something like this would be nice:
![Sidebar](https://user-images.githubusercontent.com/27908024/142183297-04a493f7-16ea-4856-934c-358d7e2a6129.png)

xou816:
That is lovely, thank you both! I generally trust @jannuary with the design stuff 😄 mock-up looks good to me! If that's okay with everyone I think that's the way to go! 
Thank you for your dedication @abegert hmu if you have any technical concerns! I think the current sidebar is a bit messy code wise

abegert:
I really like the mockup, @jannuary!
Correct me if I am wrong, but AFAIK, creating new playlists is not implemented in spot yet, so I will not be adding the 'New Playlist' item for now.

abegert:
> > What would you prefer? Moving Playlists to the bottom and then adding items for each playlist underneath?
> 
> Yeah, that sounds good. In more general thoughts about sidebar, something like this would be nice: ![Sidebar](https://user-images.githubusercontent.com/27908024/142183297-04a493f7-16ea-4856-934c-358d7e2a6129.png)

I've been making some progress on this issue. However, I couldn't find the icons you are using in the mockup. I tried to find them under /usr/share/icons. Where did you find them? And if they are gnome icons and I just overlooked them, what are their names?

[...]

PR #412, opened 2021-12-20
287 make better use of sidepanel  option to list playlists

abegert:
This implements https://github.com/xou816/spot/issues/287

I replaced the StackSideBar with a ListBox, so that we can freely modify it. I'm still using a Stack for the pages in the mainwindow and switch to them manually when the corresponding ListBoxRow is activated.

The UI looks like this now, the only change visually since the last discussion is the 'All Playlists' entry:
image

I'm happy with it, though I can think of one thing that could be worth improving: It would be good to differentiate the 'All Playlists' entry a little from the users playlists below. The only idea I've had was to add something like separators around it, but not sure if that's such a good idea.

Thoughts, @jannuary @xou816 ?


jannuary:
A design review:
- Sidebar items are now edge to edge, I don't think this is something we want?
image
FYI .navigation-sidebar on the ListBox gives proper paddings and rounded corners (though you'd need to apply ```background-color: @view_bg_color;``` in css manually)
- The sidebar seems wider than before? I think the status quo was mostly fine, I don't think playlist names are gonna be significantly long for the most part.
image
- 'All Playlists' has the same icon as every other playlist, despite not being a playlist. I think it'd help to delineate it with another icon
(though the problem there is what icon to use)

Other than that, great work! Would love to see this merged!

src/app/components/sidebar_listbox/icons/star-filled-symbolic.svg
```diff
@@ -0,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jannuary, line 1: This doesn't need to be bundled - just use `starred-symbolic`.

src/app/components/sidebar_listbox/sidebar.css
```diff
@@ -0,0 +1,3 @@
+.item_sidebar {
```
jannuary, line 1: Do we need css here?

src/app/components/sidebar_listbox/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+pub use sidebar_item::*;
+
+pub fn build_sidebar_listbox(builder: &gtk::Builder, list_store: &gio::ListStore) -> ListBox {
+    display_add_css_provider(resource!("/sidebar_listbox/sidebar.css"));
+    let listbox: gtk::ListBox = builder.object("home_listbox").unwrap();
+    listbox.bind_model(Some(list_store), move |item| {
+        let label = gtk::Label::new(Option::from(
+            item.property("title").unwrap().get::<&str>().unwrap(),
+        ));
+        let gtk_box = gtk::Box::new(gtk::Orientation::Horizontal, 10);
```
jannuary, line 18:
```suggestion
        let gtk_box = gtk::Box::new(gtk::Orientation::Horizontal, 12);
```

src/app/components/sidebar_listbox/sidebar.css
jannuary: Without `padding`, of course

abegert:
@jannuary thanks for the prompt and detailed feedback!

* I added the css class 'navigation-sidebar' and set the background-color
* I didn't deliberately change the width, it would just auto expand to accommodate the playlist with the longest name. In the case of your screenshot to the width of the second playlist from the bottom. I just made a change that limits the label to 20 characters and ellipsizes after that. I also added a tooltip so longer names can still be read somehow
* Yeah, I wasn't happy about that either, but I just found an icon that I like for it. It's a grid symbol (see the screenshot below)

The rest of your suggested changes I implemented as well.

image


Let me know what you think, I hope I understood your suggestions correctly!

jannuary:
I'm not too partial to the grid icon, might mess around with it tomorrow, but otherwise LGTM.

xou816 (commented):
Hi! This is astounding work, thank you for working on this! I have a few comments here and there (most importantly, I think we could avoid direct calls to `property` which are "flimsy" since one could easily get the wrong string here or mess up the conversion from a GSomething to a plain rust type) but for the most part the implementation is pretty good :) I'll try and keep a close eye on the PR if you need more feedback from me.

Also thanks @jannuary for helping with the review, I appreciate it!

src/app/components/navigation/factory.rs
```diff
@@ -37,9 +37,12 @@ impl ScreenFactory {
         )
     }
 
+    pub fn make_saved_playlists_model(&self) -> SavedPlaylistsModel {
```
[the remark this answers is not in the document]
abegert: I can't make it non public without some modifcation: the method is called from HomePane as well. I need the SavedPlaylistModel there, and preferred just making a new one over changing the return type of make_saved_playlists.
I'd be open for a suggestion on how to change this, it's definitely not perfect.

src/app/components/navigation/navigation.rs
```diff
@@ -11,7 +11,8 @@ pub struct Navigation {
     model: Rc<NavigationModel>,
     leaflet: libadwaita::Leaflet,
     navigation_stack: gtk::Stack,
-    home_stack_sidebar: gtk::StackSidebar,
+    home_listbox: gtk::ListBox,
```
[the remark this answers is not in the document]
abegert: I think we do. We pass it to the HomePane constructor in make_home. make_home is called from navigation, for example during push_screen.

src/app/components/sidebar_listbox/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+mod sidebar_row;
+
+use crate::app::components::sidebar_listbox::sidebar_row::SideBarRow;
+
+use crate::app::components::display_add_css_provider;
+use gtk::prelude::*;
+use gtk::ListBox;
+pub use sidebar_item::*;
+
+pub fn build_sidebar_listbox(builder: &gtk::Builder, list_store: &gio::ListStore) -> ListBox {
```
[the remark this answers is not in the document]
abegert: Do you mean a custom ListBox widget? I'm not sure whats to gain from that. I did refactor this function. It now just passes the item from the ListStore into the SideBarRow constructor, which then handles all the details.
I also wrote a custom widget for the GtkBox containing an icon and a label with certain styling, etc.

src/app/components/sidebar_listbox/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    listbox.bind_model(Some(list_store), move |item| {
+        let title_value = item.property("title").unwrap();
+        let title = Option::from(title_value.get::<&str>().unwrap());
+        let label = gtk::Label::new(title);
+        label.set_ellipsize(gtk::pango::EllipsizeMode::End);
+        label.set_max_width_chars(20);
+        let gtk_box = gtk::Box::new(gtk::Orientation::Horizontal, 12);
+        gtk_box.set_tooltip_text(title);
+        let row = SideBarRow::new(item.property("id").unwrap().get::<&str>().unwrap());
+        if item.property("grayedout").unwrap().get::<bool>().unwrap() {
```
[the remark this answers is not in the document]
abegert: Good suggestion, I had tried to do that but failed casting the item into SideBarItem. I managed to do that this time, it's much cleaner like that imo.

abegert:
Thank you for the kind words! I worked in your feedback, let me know what you think.

src/app/components/navigation/home.rs
```diff
@@ -130,7 +121,7 @@ impl HomePane {
     }
 
     pub fn connect_navigated<F: Fn() + 'static>(&self, f: F) {
-        let model = self.saved_playlists_model.box_clone();
+        let model = self.saved_playlists_model.clone();
         self.listbox
             .connect_row_activated(clone!(@weak self.stack as stack => move |_, row| {
                 let n = row.property("id").expect("Could not get id of ListBoxRow.");
```
xou816, line 127: downcast + use your new id() method maybe?

src/app/components/sidebar_listbox/mod.rs
```diff
@@ -11,34 +12,9 @@ pub use sidebar_item::*;
...
-            icon.set_icon_name(Option::from(
-                item.property("iconname").unwrap().get::<&str>().unwrap(),
-            ));
-            gtk_box.append(&icon);
-            gtk_box.append(&label);
-
-            row.set_child(Option::from(&gtk_box));
-        }
+    listbox.bind_model(Some(list_store), move |obj| {
+        let item = obj.clone().downcast::<sidebar_item::SideBarItem>().unwrap();
```
xou816, line 16: instead of clone+downcast I think you can just downcast_ref !

src/app/components/sidebar_listbox/mod.rs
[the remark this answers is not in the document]
xou816: why a custom widget -> for consistency and easy editing (ui files) but thats already cleaner, I'm ok with the current solution :)

xou816:
Think we're good to merge! Thanks a lot for this!!

PR #427, opened 2022-01-13
basic connect implementation

xou816:
image


fushinari:
Does this MR also mean that spot will now be controllable from other devices like spotifyd is?

xou816:
Nope, sorry!

xou816:
its quite a nightmare to get this right...
right now the web connect api does not expose the queue so things like shuffle dont work well
also not handling the case where you completely change the source of the playback on the connected device, say switch to another playlist, not motivated to tackle this just yet

> Does this MR also mean that spot will now be controllable from other devices like spotifyd is?

to elaborate on this: this is a very different thing, and a whole other can of worms too! I need to finish this one first, which is control other devices from Spot; controlling Spot from other devices is less interesting anyway and I'd recommend just using spotifyd/librespot if a gui is not needed anyway

xou816:
its not perfect, def not... but lets go ahead with this and see how it goes. we'll improve on it over time

PR #429, opened 2022-01-14
Update dark mode guidance

jannuary:
It has been obsolete since libadwaita anyway, but it is even more obsolete now with color scheme preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,14 +48,14 @@ Scrobbling is not supported directly by Spot. However, you can use a tool such a
 
 Similarly, Spot does not display lyrics for songs, but you can use [osdlyrics](https://github.com/osdlyrics/osdlyrics)  ([see #226](https://github.com/xou816/spot/issues/226)).
 
-### Gtk theme
+### Dark mode 
 
-Spot uses the dark theme variant by default; this can be changed using `gsettings`.
-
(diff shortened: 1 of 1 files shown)


xou816:
I have not had the chance to try out just yet! Curious to see how it works. If the system preference is set to dark or light, what does our PreferLight or PreferDark setting do exactly? Are we right to use these values?

jannuary:
> If the system preference is set to dark or light, what does our PreferLight or PreferDark setting do exactly?

System preference has three options: "Light" (not exposed in elementary or GNOME afaik), "Default" and "Dark".

With prefers-dark-mode on, "Default" or "Dark" = dark, "Light" = light;
with it off, "Light" and "Default" = light, "Dark" = dark.

Also see [Alexander's blog post](https://blogs.gnome.org/alexm/2021/10/04/dark-style-preference/)

Having PreferLight/PreferDark isn't ideal, but when we have an actual preferences window, we could have a System/Light/*Dark* tri-state toggle that maps to PreferLight/ForceLight/ForceDark

xou816:
We could make it a tristate thing already, or just have Default (ie system preference) vs PrefersDark in the meantime!

jannuary:
It could be yeah, though that also wouldn't make sense on systems without color scheme prefence.

Also, there is no "Default" from our side, just Prefer/Force Light or Dark

PR #440, opened 2022-01-16
Align album track song indices

### LINKED ISSUE #208: Track titles aren't aligned in album view
This is a minor design quirk rather than a major bug - I know I'm being fussy by reporting this. I'll counter that by saying I really like the Spot app and it has replaced the official Spotify client on all my Linux machines. Spot is much more pleasant to use.

**Describe the bug**
When looking at an album, the track titles aren't aligned with each other. When the track number jumps from '9' to '10', the titles jump slightly to the right.

**To Reproduce**
1. Open Spot
2. Find an album with more than nine tracks
3. Notice the titles aren't in line with each other.

spot-misaligned-numbers

**Expected behavior**
Track titles should be consistently left-aligned with each other. The following screenshot is the same album stored locally, displayed in Gnome Music. Notice the track titles are consistently aligned, and the track number column is slightly wider and right-aligned. Applying a minimum width to the track number and right-alignng it would probably be the neatest solution.

gnome-music-numbers-aligned

**General information:**
 - Distribution: Ubuntu 20.04
 - Installation method: Built from source and installed without Flatpak. `-Dbuildtype=release -Doffline=false --prefix="/usr"`
 - Version: 0.1.13
 - Device used: Desktop

Tested with several different GTK themes, including Adwaita, Arc, Pocillo, Qogir, and Mojave-light-alt (used in screenshots). All display the same behaviour.

keringar:
Fixes #208. Spotify appears to reset the track index every 80
songs so 1.5em (enough for two digits) seems to be good
enough.

Tested by viewing the Thirukkural album which contains 1330
tracks. Seems to work well and everything is aligned.

### DIFF
--- a/src/app/components/playlist/song.css
+++ b/src/app/components/playlist/song.css
@@ -3,6 +3,7 @@
   margin: 6px 12px;
   padding: 0;
   opacity: 1;
+  min-width: 1.5em;
 }
 
 .song__cover {
--- a/src/app/components/playlist/song.ui
+++ b/src/app/components/playlist/song.ui
@@ -19,6 +19,7 @@
           <object class="GtkLabel" id="song_index">
             <property name="label">1</property>
             <property name="sensitive">0</property>
+            <property name="halign">GTK_ALIGN_CENTER</property>
             <style>
               <class name="song__index"/>
               <class name="numeric"/>


xou816:
Oh, I had forgotten about this! Looking great! Saving this album for reference, could be useful for testing. Thanks for your contribution!

PR #446, opened 2022-01-18
added italian translation

fbabetto:
Hi,
first of all thank you for spot, I really like it!
I just added the Italian translation, I'm 99,99% satisfied by this translation, maybe in the future I'll post an update because I'm not completely satisfied by some string.

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,1 +1,1 @@
-en ru fr nl de ca cs pl es pt tr id pt-br fi eu ia nb sl ja ar
+en ru fr nl de ca cs pl es pt tr id pt-br fi eu ia nb sl ja ar it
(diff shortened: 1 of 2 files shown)


fbabetto:
Sorry, I didn't notice there was a template to follow, I'll update this PR soon :)

fbabetto:
Fixed, read to be pulled.

xou816:
Hi! Thank you for taking the time to contribute these translations! :) 

Don't worry, we don't have a template, as files get regenerated fairly often anyway. By the way, you can directly contribute on [POEditor](https://poeditor.com/join/project?hash=xfVrpQfRBM) if you want to!

As for the PR, you'll just need to add the it code to LINGUAS for this code to work, otherwise the file seems to be properly formatted! :)

xou816 (changes requested):
(missing it code in LINGUAS)

fbabetto:
Done!

xou816:
looking good :) merging, thank you very much!

PR #455, opened 2022-01-26
Add GenericName to desktop entry file

ChrisLane:
Having a generic name in the desktop entry adds some useful context.


xou816:
good call! thanks for this, merging!

Issue #413, opened 2021-12-22
Implement creating new playlists


abegert:
Hey,

I am missing the functionality to create a new playlist (I was surprised not to find an already existing issue for this).
I would like to work on this myself. UI wise, I would like to go with what @jannuary had already suggested here: https://github.com/xou816/spot/issues/287#issuecomment-971441309

Anything I should know? My plan so far is basically this:
* Add an item to the sidebar ListBox like 'New Playlist' with a '+' icon
* show some kind of small overlay right next to the 'New Playlist' item with an input box for the name and a button labeled 'Create'
* Write a new method in client.rs that POSTS to the API like this: https://developer.spotify.com/documentation/web-api/reference/#/operations/create-playlist
* call this method in the client when the create button is clicked

Any thoughts?

jannuary:
> show some kind of small overlay right next to the 'New Playlist' item with an input box for the name and a button labeled 'Create'

I'm not too sure what you mean by overlay - there are popovers and there are dialogs.

abegert:
> > show some kind of small overlay right next to the 'New Playlist' item with an input box for the name and a button labeled 'Create'
> 
> I'm not too sure what you mean by 'overlay' - there are popovers and there are dialogs.

Sorry, I was just trying to describe it in my own words and I'm not familiar with all the GTK lingo yet.
Popover looks like what I was thinking.

jannuary:
> Sorry, I was just trying to describe it in my own words and I'm not familiar with all the GTK lingo yet. Popover looks like what I was thinking.

Popovers don't usually have "Confirm" and such buttons :p
I'm thinking of something more of the lines that you create the playlist → you get sent to it → edit mode (outlined in #354) turns on, with the focus on the title entry. That way, you aren't explicitly required to name your playlist, but you can still do so quickly.

There also should be a "New Playlist" entry in the selection mode, and in that case, an unnamed playlist is transparently created, but not opened.

abegert:
> > Sorry, I was just trying to describe it in my own words and I'm not familiar with all the GTK lingo yet. Popover looks like what I was thinking.
> 
> Popovers don't usually have "Confirm" and such buttons :p I'm thinking of something more of the lines that you create the playlist → you get sent to it → edit mode (outlined in #354) turns on, with the focus on the title entry. That way, you aren't explicitly required to name your playlist, but you can still do so quickly.
> 
[...]

PR #462, opened 2022-02-03
413 implement creating new playlists

abegert:
Issue: https://github.com/xou816/spot/issues/413

Hi,

I've been working on implementing creating new playlists.
The current state is this:
There's a new sidebar entry 'New Playlist'. If you click on it, a popover appears to the right of the sidebar entry, where you can put in the name of the new playlist. You can press the 'Enter' key or click 'Create' to confirm.
A request is made to the spotify API to create the playlist with name $NAME without additional parameters (which means that spotify uses it's default: public=true, collaborative=false).
Once the playlist is created, a Notification appears saying that a new playlist with $NAME has been created.

I'm happy with the changes apart from two things:
1) We had originally discussed to include a button in the toast, which brings the user to the newly created playlist
2) After creating a new playlist, the sidebar entry below 'New Playlist' will always appear to be selected, even though another one was selected before. This will also be inconsistent with what is actually shown in the leaflet.

Regarding 1): As I understand it, we currently have the AppAction ShowNotification(String) (which is what I am using currently). It might make sense to have a more generalized ShowNotification, but I'm not sure what signature that should have. I wanted to get somebody else's opinion before working on that.
Regarding 2): My idea for how to change that was this: I wanted to save the activated row of the listbox whenever a row is activated. If we always know the row that was activated before, we can activate it again after the popover pops down. That does not work however, since ListBox::connect_row_activated expects a closure that implements 'Fn', not only 'FnMut'. We need it to be FnMut since we want the closure to mutate the previously activated row.
I haven't had a good idea how we would fix that yet.

As usual, it probably also makes sense to discuss the UI:
image

I'm looking forward to feedback and suggestions.

### DIFF
--- a/src/app/components/navigation/create_playlist.rs
+++ b/src/app/components/navigation/create_playlist.rs
@@ -1,5 +1,4 @@
 use crate::app::components::HomeModel;
-use glib::SignalHandlerId;
 use gtk::prelude::*;
 use gtk::subclass::prelude::*;
 use gtk::CompositeTemplate;
@@ -49,23 +48,25 @@ impl CreatePlaylistWidget {
     pub fn new(parent: &gtk::ListBoxRow, model: Rc<HomeModel>, user: Rc<String>) -> Self {
         let w: CreatePlaylistWidget =
             glib::Object::new(&[]).expect("Failed to create an instance of CreatePlaylistWidget");
-        w.connect_button_clicked(model, user);
+        w.connect_create(model, user);
         w.set_parent(parent);
         w.popup();
         w
     }
 
-    pub fn connect_button_clicked(
-        &self,
-        model: Rc<HomeModel>,
-        user: Rc<String>,
-    ) -> SignalHandlerId {
+    fn connect_create(&self, model: Rc<HomeModel>, user: Rc<String>) {
         let widget = imp::CreatePlaylistWidget::from_instance(self);
         let btn = widget.button.get();
         let entry = widget.entry.get();
+        entry.connect_activate(
+            clone!(@weak self as _self, @weak user, @weak model => move |entry| {
+                model.create_new_playlist(entry.text().to_string(), user.to_string());
+                _self.popdown();
+            }),
+        );
         btn.connect_clicked(clone!(@weak self as _self => move |_| {
             model.create_new_playlist(entry.text().to_string(), user.to_string());
             _self.popdown();
-        }))
+        }));
     }
 }
--- a/src/app/components/navigation/create_playlist.rs
+++ b/src/app/components/navigation/create_playlist.rs
@@ -1,8 +1,6 @@
-use crate::app::components::HomeModel;
 use gtk::prelude::*;
 use gtk::subclass::prelude::*;
 use gtk::CompositeTemplate;
-use std::rc::Rc;
 
 mod imp {
     use super::*;
@@ -45,25 +43,25 @@ glib::wrapper! {
 }
 
 impl CreatePlaylistWidget {
-    pub fn new(parent: &gtk::ListBoxRow, model: Rc<HomeModel>) -> Self {
+    pub fn new(parent: &gtk::ListBoxRow) -> Self {
         let w: CreatePlaylistWidget =
             glib::Object::new(&[]).expect("Failed to create an instance of CreatePlaylistWidget");
-        w.connect_create(model);
         w.set_parent(parent);
-        w.popup();
         w
     }
 
-    fn connect_create(&self, model: Rc<HomeModel>) {
+    pub fn connect_create<F: Clone + Fn(String) + 'static>(&self, create_fun: F) {
         let widget = imp::CreatePlaylistWidget::from_instance(self);
         let btn = widget.button.get();
         let entry = widget.entry.get();
-        entry.connect_activate(clone!(@weak self as _self, @weak model => move |entry| {
-            model.create_new_playlist(entry.text().to_string());
-            _self.popdown();
-        }));
+        entry.connect_activate(
+            clone!(@weak self as _self @strong create_fun => move |entry| {
+                create_fun(entry.text().to_string());
+                _self.popdown();
+            }),
+        );
         btn.connect_clicked(clone!(@weak self as _self => move |_| {
-            model.create_new_playlist(entry.text().to_string());
+            create_fun(entry.text().to_string());
             _self.popdown();
         }));
     }
--- a/src/app/components/navigation/create_playlist.rs
+++ b/src/app/components/navigation/create_playlist.rs
@@ -1,0 +1,71 @@
+use crate::app::components::HomeModel;
+use glib::SignalHandlerId;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use gtk::CompositeTemplate;
+use std::rc::Rc;
+
+mod imp {
+    use super::*;
+
+    #[derive(Debug, Default, CompositeTemplate)]
+    #[template(resource = "/dev/alextren/Spot/components/create_playlist.ui")]
+    pub struct CreatePlaylistWidget {
+        #[template_child]
+        pub label: TemplateChild<gtk::Label>,
+
+        #[template_child]
+        pub entry: TemplateChild<gtk::Entry>,
+
+        #[template_child]
+        pub button: TemplateChild<gtk::Button>,
+    }
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for CreatePlaylistWidget {
+        const NAME: &'static str = "CreatePlaylistWidget";
+        type Type = super::CreatePlaylistWidget;
+        type ParentType = gtk::Popover;
+
+        fn class_init(klass: &mut Self::Class) {
+            Self::bind_template(klass);
+        }
+
+        fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
+            obj.init_template();
+        }
+    }
+
+    impl ObjectImpl for CreatePlaylistWidget {}
+    impl WidgetImpl for CreatePlaylistWidget {}
+    impl PopoverImpl for CreatePlaylistWidget {}
+}
+
+glib::wrapper! {
+    pub struct CreatePlaylistWidget(ObjectSubclass<imp::CreatePlaylistWidget>) @extends gtk::Widget, gtk::Popover;
+}
+
+impl CreatePlaylistWidget {
+    pub fn new(parent: &gtk::ListBoxRow, model: Rc<HomeModel>, user: Rc<String>) -> Self {
+        let w: CreatePlaylistWidget =
+            glib::Object::new(&[]).expect("Failed to create an instance of CreatePlaylistWidget");
+        w.connect_button_clicked(model, user);
+        w.set_parent(parent);
+        w.popup();
+        w
+    }
+
+    pub fn connect_button_clicked(
+        &self,
+        model: Rc<HomeModel>,
+        user: Rc<String>,
+    ) -> SignalHandlerId {
+        let widget = imp::CreatePlaylistWidget::from_instance(self);
+        let btn = widget.button.get();
+        let entry = widget.entry.get();
+        btn.connect_clicked(clone!(@weak self as _self => move |_| {
+            model.create_new_playlist(entry.text().to_string(), user.to_string());
+            _self.popdown();
+        }))
+    }
+}
--- a/src/app/components/navigation/create_playlist.rs
+++ b/src/app/components/navigation/create_playlist.rs
@@ -45,27 +45,25 @@ glib::wrapper! {
 }
 
 impl CreatePlaylistWidget {
-    pub fn new(parent: &gtk::ListBoxRow, model: Rc<HomeModel>, user: Rc<String>) -> Self {
+    pub fn new(parent: &gtk::ListBoxRow, model: Rc<HomeModel>) -> Self {
         let w: CreatePlaylistWidget =
             glib::Object::new(&[]).expect("Failed to create an instance of CreatePlaylistWidget");
-        w.connect_create(model, user);
+        w.connect_create(model);
         w.set_parent(parent);
         w.popup();
         w
     }
 
-    fn connect_create(&self, model: Rc<HomeModel>, user: Rc<String>) {
+    fn connect_create(&self, model: Rc<HomeModel>) {
         let widget = imp::CreatePlaylistWidget::from_instance(self);
         let btn = widget.button.get();
         let entry = widget.entry.get();
-        entry.connect_activate(
-            clone!(@weak self as _self, @weak user, @weak model => move |entry| {
-                model.create_new_playlist(entry.text().to_string(), user.to_string());
-                _self.popdown();
-            }),
-        );
+        entry.connect_activate(clone!(@weak self as _self, @weak model => move |entry| {
+            model.create_new_playlist(entry.text().to_string());
+            _self.popdown();
+        }));
         btn.connect_clicked(clone!(@weak self as _self => move |_| {
-            model.create_new_playlist(entry.text().to_string(), user.to_string());
+            model.create_new_playlist(entry.text().to_string());
             _self.popdown();
         }));
     }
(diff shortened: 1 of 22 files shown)


xou816:
Hi there! Off to a good start I see!!

I am no expert with the aesthetics part, but one suggestion I would have is to have "Name" be a placeholder to the input so that we can be rid of a label and the extra space it takes.

As for your other two points:

1. Try and find something that works for you, as a separate action, we might not need to generalize just yet :) if I were to do it, I'd suggest something like ShowActionNotification { message: String, action_label: String, action: Box<AppAction> } if you see what I mean]

2.  I am not entirely sure what you are trying to achieve, but I doubt you actually need FnMut -- Gtk widgets usually implement interior mutability in some way and a plain ref is enough to mutate them. Could you give some details about what you're trying to do? But really what we should be doing is apply the usual pattern here (so you don´t have to worry about cloning a bunch of widgets inside closures), and have some model that reads the app state to know where we stand navigation-wise. I understand it might not be so easy atm since I never took the time to properly manage the sidebar state as part of the app state. I could handle some of that boilerplate over the weekend if you feel this might help!

Nice work otherwise, looking forward to see it in action :)

jannuary:
> 1. ShowActionNotification { message: String, action_label: String, action: Box }

Box? Toasts take a GAction: [Adw.Toast](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.Toast.html#actions)

xou816:
Ooops, I missed a word -- I meant Box<AppAction>, as in, this action wraps another action that would be triggered within a Toast indeed

abegert:
> Hi there! Off to a good start I see!!
> 
> I am no expert with the aesthetics part, but one suggestion I would have is to have "Name" be a placeholder to the input so that we can be rid of a label and the extra space it takes.
> 
> As for your other two points:
> 
>     1. Try and find something that works for you, as a separate action, we might not need to generalize just yet :) if I were to do it, I'd suggest something like ShowActionNotification { message: String, action_label: String, action: Box } if you see what I mean]
> 
>     2. I am not entirely sure what you are trying to achieve, but I doubt you actually need FnMut -- Gtk widgets usually implement interior mutability in some way and a plain ref is enough to mutate them. Could you give some details about what you're trying to do? But really what we should be doing is apply the usual pattern here (so you don´t have to worry about cloning a bunch of widgets inside closures), and have some model that reads the app state to know where we stand navigation-wise. I understand it might not be so easy atm since I never took the time to properly manage the sidebar state as part of the app state. I could handle some of that boilerplate over the weekend if you feel this might help!
> 
> 
> Nice work otherwise, looking forward to see it in action :)

Good idea about the label. I got rid of the label and put in a placeholder text for the Entry.
image

Regarding 1):
Sorry, I can't quite follow. I thought you meant something like this:
ShowActionNotification(String, String, Box<gio::Action>)
But that won't work, since we can't send actions across threads because they contain some mutable references that are not thread safe.

Regarding 2):
I looked into this myself over the weekend. I had actually already introduced a 'HomeModel' and am using that to call the dispatcher. I haven't made much progress on using it to access the app state etc.
A little help here would be appreciated, but I understand if you can't make the time for it.

Thanks for the help so far!

jannuary:
Having the title be in placeholder presents bit of an accessibility issue, so I'd just advise copying what Nautilus does here:
Screenshot from 2022-02-07 21-02-36

xou816:
I will have a proper look a bit later but I meant `Box<AppAction>`... sorry, just realized GitHub gobbles words if you use chevrons in plain text... Otherwise you can just have a PlaylistCreated event that's handled by the notification component, that's fine too!

abegert:
> Having the title be in placeholder presents bit of an accessibility issue, so I'd just advise copying what Nautilus does here: Screenshot from 2022-02-07 21-02-36

Copying the .ui file from Nautilus with only renaming always gives me something very odd looking:
image
I'm going to go back to what I had before with the label, see above.

I also worked on the Toast for a while but got stuck:
The button appears to be deactivated:
image

I haven't figured out why, maybe one of you has an idea?

xou816:
Looks like you haven't defined the `gio::Action` you mean to trigger I'd say!

abegert:
> Looks like you haven't defined the `gio::Action` you mean to trigger I'd say!

I don't think that's it, I have defined it, but probably not quite correctly. If I set action_name on the Toast to an action that doesn't exist, it behaves exactly the same. I'll keep looking into it.

xou816:
oh okay saw it now, I think you have to call it as app.open_playlist (you can also use the Gtk inspector to see actions)

abegert:
> oh okay saw it now, I think you have to call it as app.open_playlist (you can also use the Gtk inspector to see actions)

Oh wow, thanks, that works. I could have sworn I had already tried that. Thanks for the tip about GTK inspector too, I hadn't heard of that before, it's pretty neat.

I also just rebased on /development, hence all the new commits.

I think that mostly leaves the issue of making sure that the state of the app is consistent with the Sidebar listbox, which I'll be working on.

abegert:
Hi there, sorry about the hiatus, I was busy with other things.

I went into the direction you mentioned here under 2.:

> Hi there! Off to a good start I see!!
> 
> I am no expert with the aesthetics part, but one suggestion I would have is to have "Name" be a placeholder to the input so that we can be rid of a label and the extra space it takes.
> 
> As for your other two points:
> 
>     1. Try and find something that works for you, as a separate action, we might not need to generalize just yet :) if I were to do it, I'd suggest something like `ShowActionNotification { message: String, action_label: String, action: Box<AppAction> }` if you see what I mean]
> 
>     2. I am not entirely sure what you are trying to achieve, but I doubt you actually need FnMut -- Gtk widgets usually implement interior mutability in some way and a plain ref is enough to mutate them. Could you give some details about what you're trying to do? But really what we should be doing is apply the usual pattern here (so you don´t have to worry about cloning a bunch of widgets inside closures), and have some model that reads the app state to know where we stand navigation-wise. I understand it might not be so easy atm since I never took the time to properly manage the sidebar state as part of the app state. I could handle some of that boilerplate over the weekend if you feel this might help!
> 
> 
> Nice work otherwise, looking forward to see it in action :)

Basically, I am saving the currently and previously selected row in a SidebarState class and then, when 'All Playlists' receives a 'select' signal, I check if it is selected even though I don't have it as selected in the app state. If that's the case, then I reselect the row that I have saved as selected in the app state.

This is sort of hacky but it's the best I could come up with.
Before, I selected the 'correct' row in the popdown signal of the Popover, which would result in the correct row being selected for a split second, then 'All Playlists' would be selected. So that was simply to soon.

The one remaining issue (apart from design decisions that are probably also worth discussing) is this:
When I click 'New Playlist', then close the popover, the ListBoxRow underneath it ('All Playlists') is not selected (because I am overriding it with what  I described above) but it does receive this green outline. I don't know what that is, what to call it (so I can research how to stop it) and I don't think there's another scenario where it appears. Some help in that direction would be much appreciated!
image
So for this screenshot, I had 'Now Playing' selected, then clicked 'New Playlist'. After the Popover pops down, the correct row is selected again, it just has this green line around it.

src/app/components/navigation/home.rs
```diff
@@ -117,22 +147,51 @@ impl HomePane {
...
         }
     }
 
-    pub fn connect_navigated<F: Fn() + 'static>(&self, f: F) {
-        let model = self.saved_playlists_model.clone();
-        self.listbox
-            .connect_row_activated(clone!(@weak self.stack as stack => move |_, row| {
+    pub fn connect_navigated(&mut self) {
+        let playlist_model = self.saved_playlists_model.clone();
+        let f = self.listbox_fun.clone();
```
xou816, line 159: is it not possible to keep the old way, am I missing something? (not storing the fun, just passing it as an arg to connect_navigated)

src/app/components/navigation/home.rs
```diff
@@ -169,6 +228,27 @@ impl EventListener for HomePane {
             AppEvent::BrowserEvent(BrowserEvent::SavedPlaylistsUpdated) => {
                 self.update_playlists_in_sidebar();
             }
+            AppEvent::LoginEvent(LoginEvent::LoginStarted(event)) => {
```
xou816, line 231: a lot of the complexity here (storing handlers, unbinding rebinding, etc) could be avoided I think! your model should be able to get the user id from the state instead of this, and then call spotify

src/app/components/navigation/create_playlist.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn new(parent: &gtk::ListBoxRow, model: Rc<HomeModel>, user: Rc<String>) -> Self {
+        let w: CreatePlaylistWidget =
+            glib::Object::new(&[]).expect("Failed to create an instance of CreatePlaylistWidget");
+        w.connect_create(model, user);
+        w.set_parent(parent);
+        w.popup();
+        w
+    }
+
+    fn connect_create(&self, model: Rc<HomeModel>, user: Rc<String>) {
```
xou816, line 57: i'd prefer to follow the usual pattern of passing an impl Fn here, so that the widget does not depend on some specific model, but rather just exposes a way to handle its activation

(also see other comments, user can be removed here)

src/app/components/navigation/home_model.rs
```diff
@@ -0,0 +1,66 @@
...
+            .call_spotify_and_dispatch_many(move || async move {
+                api.create_new_playlist(name.as_str(), user_id.as_str())
+                    .await
+                    .map(|p| {
+                        vec![
+                            BrowserAction::PrependPlaylistsContent(vec![p.clone()]).into(),
+                            AppAction::ShowPlaylistCreatedNotification(
+                                // translators: This is a notification that pop ups when a new playlist is created. It includes the name of that playlist.
+                                gettext("New playlist created."),
+                                name,
```
xou816, line 33: feels weird to have the name as the notification action label -- something like "View playlist" or "Open" would do :)

src/app/components/navigation/home_model.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.dispatcher
+            .call_spotify_and_dispatch_many(move || async move {
+                api.create_new_playlist(name.as_str(), user_id.as_str())
+                    .await
+                    .map(|p| {
+                        vec![
+                            BrowserAction::PrependPlaylistsContent(vec![p.clone()]).into(),
+                            AppAction::ShowPlaylistCreatedNotification(
+                                // translators: This is a notification that pop ups when a new playlist is created. It includes the name of that playlist.
+                                gettext("New playlist created."),
```
xou816, line 32: (nitpick) could be moved to notification, if this is meant never to change

xou816 (changes requested):
Hi! First of all: no worries at all!! I am not being very active either, and no one pressures you, hopefully :) thanks for working on this!

The whole selecting the correct row business seems very complicated. Do you have a previous commit reference I could checkout to help you investigate this split-second selection thing? Maybe there is something we can do that would avoid all this complex state management to solve it? Maybe not, but let's double check! I wonder what's the behavior if we make that New Playlist row not selectable for instance?

Otherwise, I added a few more comments on the other changes. Give me a heads up if something does not make sense! 

Thanks again!

abegert:
> Hi! First of all: no worries at all!! I am not being very active either, and no one pressures you, hopefully :) thanks for working on this!
> 
> The whole selecting the correct row business seems very complicated. Do you have a previous commit reference I could checkout to help you investigate this split-second selection thing? Maybe there is something we can do that would avoid all this complex state management to solve it? Maybe not, but let's double check! I wonder what's the behavior if we make that New Playlist row not selectable for instance?
> 
> Otherwise, I added a few more comments on the other changes. Give me a heads up if something does not make sense!
> 
> Thanks again!

Regarding the split-second selection: I will whip something up and post a link to that later. I had tried making 'New Playlist' non-selectable, that didn't help though, it just changed the styling when 'New Playlist' was activated IIRC.

src/app/components/navigation/home.rs
abegert: Very helpful, thanks. I just started discovering how the state worked a few days ago, that's the only reason why I didn't to it like that in the first place!

src/app/components/navigation/create_playlist.rs
abegert: Helpful, and changed

src/app/components/navigation/home_model.rs
abegert: That also works, changed it!

src/app/components/navigation/home_model.rs
abegert: True, I think that was left over from when I had the ambition of creating a more general notification. Changed it!

abegert:
Just to re-iterate: What I was talking about with the split second selection was this: If I select the correct row on close of the new playlist popover, it will shortly select the previous/correct item, then go back.
I re-implemented what I had and pushed it here:
https://github.com/abegert/spot/commit/69fd2c121b7041a1feee126a0c2a571eaae44817

But that only slightly simplifies things, we still need to save the state.

abegert:
And one last thing: Do you have a tip what the green outline thing is called? I would also like to get rid of that. (see my comment https://github.com/xou816/spot/pull/462#issuecomment-1055314883)

xou816:
Right! State management is good, but it seems fairly convoluted at the moment. Ideally it should not allow storing invalid states (eg New playlist being selected) and avoid storing purely technical details when possible.

So in our case I´d expect the state to only actually store the name of one of the selectable rows and no other, this way it can be read and restored at any point and yield something correct

abegert:
> Right! State management is good, but it seems fairly convoluted at the moment. Ideally it should not allow storing invalid states (eg New playlist being selected) and avoid storing purely technical details when possible.
> 
> So in our case I´d expect the state to only actually store the name of one of the selectable rows and no other, this way it can be read and restored at any point and yield something correct

That sounds reasonable but I don't know how to implement it that way with the current approach. I just slightly rewrote it to update the state when the select signal is emited (only if the row is not the new playlist one) and then reselect the currently selected one. So that would override the selections that were done implicitly, such as the select of all playlists after creating a new one.
However, that approach doesn't work, it will always go back to the previously selected row. I think this is because when the reselect is called, the state has not been updated, since that happens asynchronously.

Hope I am explaining it well enough. So I can't think of an approach that does not involve keeping both the previously and the currently selected row.

xou816:
I think I don't have a good grasp on what's happening, I'll run it for myself asap. Thanks!

xou816:
Okay so upon further inspection I think I see what's happening (well not fully)

When we exit the popover, keyboard focus is transferred back to the main window as it should, and seems to be moved to the next focusable element that it finds (the All Playlists row), I don't know why it focuses and triggers selection too, though

Not so easy to fix however we handle it!

I guess one possible workaround would be to change the widget hierarchy so that the next focusable element when coming back from the popover would be something else, maybe the button that shows the popover itself. I wonder if having something like 

box
|_ popover
|_ button

would work.

abegert:
> Okay so upon further inspection I think I see what's happening (well not fully)
> 
> When we exit the popover, keyboard focus is transferred back to the main window as it should, and seems to be moved to the next focusable element that it finds (the All Playlists row), I don't know why it focuses and triggers selection too, though
> 
> Not so easy to fix however we handle it!
> 
> I guess one possible workaround would be to change the widget hierarchy so that the next focusable element when coming back from the popover would be something else, maybe the button that shows the popover itself. I wonder if having something like
> 
> box |_ popover |_ button
> 
> would work.

Thanks for the tip, I'm going to look into it in detail soon. I also saw that I actually hadn't set a parent for all widgets. Could also be interesting to set the parent of popover to the listbox and then utilize set_pointing_to to have it pointing at the correct spot. Thanks!

xou816:
if this proves too much right now, we can probably work with that for now, i want to do some work in this area anyway so there'll be other occasions to fix this little glitch!

abegert:
Hey, so I have been trying to get it to work by setting the parent of the popover to listbox, but I can't stop it from segfaulting in `gtk_list_box_get_next_visible`. I'm not sure what that's about, but I think the approach of getting rid of this appstate stuff I wrote most recently and try to get it working with correctly setup parents and attributes for the ListBoxRows is more promising and less hacky.

Regarding how we should proceed: Totally up to you.

xou816:
hey sorry I forgot about this being the original PR! I continued your work on a different branch as I was doing some refactoring, and I merged it recently, thanks a lot @abegert !! if there's something you'd like to see changed feel free to open up issues/PRs

Issue #142, opened 2021-03-14
Add GUI for GSettings


xou816:

tmpr:
I might right away look into this. I will try to add a "Theme: {Light, Dark, System} " in the options.

xou816:
Thanks! :) if you need any pointers on where to put stuff or whatever, feel free to ask!

sei0o:
I'm thinking of working on this issue. [Shortwave](https://gitlab.gnome.org/wrott1/Shortwave) (internet radio app) [uses `AdwPreferencesWindow`](https://gitlab.gnome.org/wrott1/Shortwave/-/blob/master/data/gtk/settings_window.ui)  and I suppose we can make use of this.

xou816:
Definitely the way to go :) thanks sei0o !!

sei0o:
This could be closed now? :)

xou816:
Oh indeed :) thanks for the heads up!!

pesader:
Would it be possible to add a "Follow system theme" option to the settings? It could change the gsettings `prefers-dark-theme` to `false`, as you mentioned [here](https://github.com/xou816/spot/issues/187#issuecomment-816530939).

I got that Spot [stays dark intentionally](https://github.com/xou816/spot/issues/209#issuecomment-860180537), but I'd love to have light/dark consistency especially now that Dark Mode is officially a thing on GNOME 42 :)

And thanks a lot for developing Spot! It is amazing to have a good looking open source client for my favorite music streaming service :musical_note::heart:

xou816:
Hi! Yes, this sounds reasonable. Could you open a dedicated issue for this? Thanks :)

PR #467, opened 2022-02-09
GUI for GSettings

sei0o:
I'm working on #142 and trying to create the UI first (see `components/settings/`) then to integrate it with some structs in `src/settings.rs`.

Currently I am stuck at how to handle the `AdwPreferencesWindow`. I have set up XML file with some widgets, added a button in the user menu, but an empty window is shown: 

image


xou816:
Hi! Nice, thank you for working on this!

Haven't played with the Preferences stuff yet, but looks like you might need to have a PreferencesPage as a child of that PreferencesWindow first

(...also, the `<child>` tag should only have one `<object>` child!)

jannuary:
You need to have the settings.ui file defined in the gresource schema, like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<gresources>
  <gresource prefix="/dev/alextren/Spot">
...
    <file alias="components/settings.ui">app/components/settings/settings.ui</file>
...
  </gresource>
</gresources>
```

sei0o:
It is working after applying your advice and some other tweaks. Thanks!
image

sei0o:
So now the settings window is working and the updated GSettings properties are loaded on next startup. 

I would like to make they are reloaded during the execution of Spot, as many applications do. Based on my understanding, players and some other components make use of `SpotSettings`. On startup,  the `SpotSettings` is `clone`d & passed down on those components. Therefore they are not synchronized with the original one or GSettings. 

My idea is to put a `SpotSettings` in the `AppState` and configure necessary signals/actions to make reloading happen while the app is running (the design doc is quite helpful!). What do you think?

xou816:
Neat!

Given that most if not all settings require a restart, it is not that big of a deal if we don't manage these settings through the shared AppState -- but you can absolutely do that if you want :)

As for persisting those settings, there's an example of that in settings.rs -- you could do it from pretty much anywhere I think.

sei0o:
I've just pushed the latest commit, which allows us to apply new settings without restarting. The problem, however, is that the window cannot be closed twice. This bug has been there since I created an empty PreferencesWindow and is pretty weird...

jannuary:
> I've just pushed the latest commit, which allows us to apply new settings without restarting. The problem, however, is that the window cannot be closed twice. This bug has been there since I created an empty PreferencesWindow and is pretty weird... 

It's a known issue, you need to have hide-on-close enabled on the PreferencesWindow.

src/app/components/user_menu/user_menu.rs
```diff
@@ -53,6 +62,8 @@ impl UserMenu {
     fn update_menu(&self) {
         let menu = gio::Menu::new();
         // translators: This is a menu entry.
+        menu.append(Some(&gettext("Settings")), Some("menu.settings"));
```
jannuary, line 65: ~~Settings~~ Preferences

src/app/components/settings/settings.ui
```diff
@@ -0,0 +1,86 @@
...
+                      <item>PulseAudio</item>
+                      <item>ALSA</item>
+                    </items>
+                  </object>
+                </property>
+              </object>
+            </child>
+            <child>
+              <object class="AdwActionRow">
+                <property name="title" translatable="yes" comments="Title for an item in preferences">ALSA Device</property>
```
jannuary, line 28: This should be hidden if ALSA isn't used.

src/app/components/settings/settings.ui
```diff
@@ -0,0 +1,86 @@
...
+                <property name="subtitle" translatable="yes" comments="Description for the item (ALSA Device) in preferences">Applied only if audio backend is ALSA</property>
+                <child>
+                  <object class="GtkEntry" id="alsa_device"></object>
+                </child>
+              </object>
+            </child>
+            <child>
+              <object class="AdwComboRow" id="player_bitrate">
+                <property name="title" translatable="yes" comments="Title for an item in preferences">Player bitrate</property>
```
jannuary, line 37: I think it makes more sense to expose this setting as something like "Audio Quality" with Low/Medium/High options, just like Spotify itself does.

src/app/components/settings/settings.ui
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <requires lib="gtk" version="4.0" />
+  <requires lib="libadwaita" version="1.0" />
+  <template class="SettingsWindow" parent="AdwPreferencesWindow">
+    <property name="default-width">600</property>
+    <property name="hide-on-close">1</property>
```
jannuary, line 7: We don't have nearly enough options to justify it, so disable the search:
```suggestion
    <property name="hide-on-close">1</property>
    <property name="search-enabled">0</property>
```

src/app/components/settings/settings.ui
```diff
@@ -0,0 +1,86 @@
...
+                  </object>
+                </property>
+              </object>
+            </child>
+            <child>
+              <object class="AdwActionRow">
+                <property name="title" translatable="yes" comments="Title for an item in preferences">ALSA Device</property>
+                <property name="subtitle" translatable="yes" comments="Description for the item (ALSA Device) in preferences">Applied only if audio backend is ALSA</property>
+                <child>
+                  <object class="GtkEntry" id="alsa_device"></object>
```
jannuary, line 31:
```suggestion
                  <object class="GtkEntry" id="alsa_device">
                      <property name="valign">center</property>
                  </object>
```
Otherwise it looks weirdly stretched.

src/app/components/settings/settings.ui
```diff
@@ -0,0 +1,86 @@
...
+        </child>
+        <child>
+          <object class="AdwPreferencesGroup">
+            <property name="title" translatable="yes" comments="Header for a group of preference items regarding network">Network</property>
+            <child>
+              <object class="AdwActionRow">
+                <property name="title" translatable="yes" comments="Title for an item in preferences">Access Point Port</property>
+                <property name="subtitle" translatable="yes" comments="Longer description for an item (Access Point Port) in preferences">Port used for connections to Spotify's Access Point. Set to 0 if any port is fine.</property>
+                <child>
+                  <object class="GtkEntry" id="ap_port"></object>
```
jannuary, line 77: Same `valign` as the other entry.

jannuary (changes requested):

xou816:
Hi! been a little busy but I'll have a look :)

src/app/state/settings_state.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl From<SettingsEvent> for AppEvent {
+    fn from(settings_event: SettingsEvent) -> Self {
+        Self::SettingsEvent(settings_event)
+    }
+}
+
+#[derive(Default)]
+pub struct SettingsState {}
```
xou816, line 27: I think  you could store the Settings here as mentioned at some point :)

That way, the action could have no payload with it (just `ChangePlayerSettings`) and whenever that action is dispatched, we just do `self.settings = SpotSettings::new_from_gsettings()` since it's already synced probably? 

that way you don't have to do the mapping from the Adwaita widgets to SpotSettings everytime

src/app/components/settings/settings.rs
```diff
@@ -0,0 +1,271 @@
...
+            settings_window,
+            settings,
+        }
+    }
+
+    fn window(&self) -> &libadwaita::Window {
+        let pref_window = self
+            .settings_window
+            .upcast_ref::<libadwaita::PreferencesWindow>();
+        pref_window.upcast_ref::<libadwaita::Window>()
```
xou816, line 259: think you can upcast_ref to a window in one go, no need to cast to a preferences window first

src/app/components/settings/settings.rs
```diff
@@ -0,0 +1,271 @@
...
+            });
+        });
+    }
+
+    fn connect_close<F>(&self, on_close: F)
+    where
+        F: Fn(SpotSettings) + 'static,
+    {
+        let pref_window = self.upcast_ref::<libadwaita::PreferencesWindow>();
+        let window = pref_window.upcast_ref::<libadwaita::Window>();
```
xou816, line 182: think you can upcast_ref to a window in one go, no need to cast to a preferences window first

src/player/player.rs
```diff
@@ -51,13 +51,13 @@ pub trait SpotifyPlayerDelegate {
     fn notify_playback_state(&self, position: u32);
 }
 
-#[derive(Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
xou816, line 54: I think (I need to double check) that the derived PartialEq impl won't check for equality on the nested String here, so there would be Alsa("foo") == Alsa("bar")

xou816 (commented):
nice work!! I'll try to be more reactive to questions and feedback as you go through the review!

src/app/components/settings/settings.ui
sei0o: Agreed. Current bitrate options (96/160/320 kbps) correspond to [Normal/High/"Very high" in Spotify](https://support.spotify.com/us/article/audio-quality/), so it would be better to reflect that in our preferences?

src/player/player.rs
sei0o: Double checked on [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2eb03c1532fce3300c3aa1e7a32f6d01). `derive` handles it well I suppose :)

src/app/state/settings_state.rs
sei0o: Makes sense. Moved SpotSettings into SettingsState and SettingsWindow use it whenever possible, while there are still some redundant `SpotSettings::new_from_gsettings()` calls from the player and `update_with` handlers for SettingsState.

xou816: I am nitpicking a bit, but I think one action is enough here, and the state should decide whether to return no event or `vec![SettingsEvent::PlayerSettingsChanged]` if the player part of the settings have changed

sei0o: `set_player_settings` still has to emit `PlaybackAction::Stop` to stop playing (otherwise there will be some inconsistency in UI), so the two same `if`s are put in `update_with` and `Settings::new`. It's not a nitpick because I had the same concern, just thought of avoiding the duplicate ifs at first.

src/app/state/settings_state.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+impl UpdatableState for SettingsState {
+    type Action = SettingsAction;
+    type Event = AppEvent;
+
+    fn update_with(&mut self, action: std::borrow::Cow<Self::Action>) -> Vec<Self::Event> {
+        match action.into_owned() {
+            SettingsAction::ChangeSettings => {
+                let old_settings = self.settings.clone();
```
xou816, line 40:
```diff
     fn update_with(&mut self, action: std::borrow::Cow<Self::Action>) -> Vec<Self::Event> {
         match action.into_owned() {
             SettingsAction::ChangeSettings => {
-                let old_settings = self.settings.clone();
-                self.settings = SpotSettings::new_from_gsettings().unwrap_or_default();
-                if self.settings.player_settings == old_settings.player_settings {
-                    vec![]
-                } else {
+                let old_settings = &self.settings;
+                let new_settings = SpotSettings::new_from_gsettings().unwrap_or_default();
+                let player_settings_changed =
+                    new_settings.player_settings != old_settings.player_settings;
+                self.settings = new_settings;
+                if player_settings_changed {
                     vec![SettingsEvent::PlayerSettingsChanged.into()]
+                } else {
+                    vec![]
                 }
             }
         }
```

(you can avoid the extra clone here)

src/app/state/settings_state.rs
xou816: Oh you're right absolutely!

Diegovsky:
Great work! I can't wait for this to me merged :)

src/app/state/settings_state.rs
sei0o: Fixed in the latest commit :)

xou816:
Thank you for your work and patience through the review! Merging this :)

sei0o:
Learned a lot from this PR... I really appreciate reviews & advice!

PR #480, opened 2022-03-06
WIP: Use GtkGridView for albums

martymichal:
An attempt at moving away from FlowBox to GridView for album lists. Has some rough edges (e.g., layout, stylesheet) but is usable enough. Some commits might be needed to be removed as GridView is not usable everywhere.

Is this the way to go?


xou816:
Hi! Sorry for the slow response. This sounds like a nice improvement! I'll check out that branch and have a look. Thnaks for working on this!

Diegovsky:
@martymichal can you rebase against maintenance please?

Issue #481, opened 2022-03-11
New icon?


bertob:
The current icon isn't really in line with the HIG, would you be interested in help with a new one?

Here's a first attempt, but I'd be happy to try something else if you have ideas/suggestions.

![image](https://user-images.githubusercontent.com/1908896/157866025-bf5de39c-3cd0-4b41-acc4-0f8282d5e153.png)

xou816:
Besides HIG, this is a vast improvement over the current icon I came up with! thank you for this.

I like the design you came up with, I don't have much to say, looks really good to me!

bertob:
Cool, happy to hear that! I'll add a symbolic and open an MR.

Barraguesh:
This is pretty cool! I would say it's only lacking the Spotify black waves inside the green part.

jannuary:
Prolly should be closed :p

PR #483, opened 2022-03-13
icon: new spotlight metaphor, add symbolic, add source file

bertob:
This adds a new app icon, as discussed in #481. I just replaced the assets in this MR, so the build system needs adjusting in a followup.

image


xou816:
Hi! Again, this looks so much better than the existing icon, thanks a lot! I like how the symbolic icon gives another take on that metaphor too :)

As for the build system adaptation: do you mean we should rasterize the svg as part of the build process? I'm guessing standard tools exist for that?

bertob:
No, you don't need to rasterize it, standard practice is to just use the SVG.

xou816:
Oh okay so it's just a matter of putting files in the right folders then I suppose :)

bertob:
I think the file paths in the data folder don't really matter, it's mostly a matter of adjusting the gresource/build system stuff. Not really my area of expertise though :)

Toorero:
Just execute this in the git root, and it works flawlessly no need to adjust the build system.
```
mkdir -p data/hicolor/scalable/apps
mv data/dev.alextren.Spot.svg data/hicolor/scalable/apps
```

The other option would be (as you suggested) to adjust the build system. For inspiration, you could use [Newsflash](https://gitlab.com/news-flash/news_flash_gtk/-/tree/master/data/icons), but I think that's unnecessary. Just move it.

xou816:
Hi! Yes, indeed, thanks @Toorero !

The build system already installs the hicolor directory: https://github.com/xou816/spot/blob/development/data/meson.build#L6

So indeed moving data/dev.alextren.Spot.svg to  data/hicolor/scalable/apps should be enough

And the symbolic icon should be in data/hicolor/symbolic/apps I think

xou816:
I  took the liberty to move the files as mentioned, and merged :) thanks a lot!

PR #492, opened 2022-03-15
Updates the translation files

Toorero:
I just updated the translation files and added myself as a translator for the German language. I also fixed a translation bug in Interlingua.


xou816:
Hi! Thanks a lot for that :) would you mind using POEditor for these instead? (see README) You can still open a PR here for the TRANSLATORS file or other support files that would need updating, but it's easier for me to import all translations at once from POEditor every now and then, if that's okay with you?

PR #526, opened 2022-05-23
Mark as single main window app

nicolasfella:
That way desktop environments know not to offer launching a second instance


xou816 (approved):

xou816:
Good call! thanks for your contribution :)

PR #529, opened 2022-06-11
Add volume slider

### LINKED ISSUE #77: App volume control
It would be very convenient to have a slider integrated in the app to control the app volume independently from the general volume (I often lower it to prevent missing out other desktop/apps sounds, but I have to go through the audio settings - using Linux Mint 20).

P.S. Love the project! Merci

Diegovsky:
Hopefully we can fully close #77 now


Diegovsky:
image

This is different from the design I showed off at #77 earlier today. Mainly, the volume slider is fixed size and embedded right into the control widget. Thoughts?

OKPersson:
I would try and move it a bit to the left so that the music icon and the volume slider are on each side of the playback controls with. But now when I looked at the app when it's playing music I realize that there's more info than in your screenshot, The playback timer is around where the slider should be imho. 
Although I have no idea where the time remaining should be placed instead.

Diegovsky:
@OKPersson I agree with you. I sampled some designs from other media players, such as Amberol and Spotify and they both have two "rows" of controls so everything fits together

Diegovsky:
> Cool, I think that it would be best located roughly the same as the default Spotify app. However I don't know how that would work with smaller layouts such as the mobile view.

Answering here for the sake of relevance: Yes, I would love that too. I'm a bit cautious of changing the layout, though, so let's wait for other contributors to weigh in :)

Diegovsky:
image

Ok, so this is my second iteration of the design. Since the lower bar is quite a bit bigger, I had to also scale the album image to fit better, though I think it's too big

Also, I want to put a speaker icon before and after the slider for better looks

Zaedus:
I have an idea for the design. Personally, I don't think that the volume should be visible all the time. Take for instance Celluloid, the GTK mpv frontend, which uses a popout volume slider.
2022-06-14T14:02:06,300999579-05:00

If it needs to be shown, maybe we go for a slider which is right oriented like in the original Spotify client?
image

Maybe we could do some kind of revealer which shows the Spotify-like volume slider when you hover over the volume icon? All just ideas, but tell me what you think!

Diegovsky:
I would prefer to mix both ideas and use a spotify-like scheme for desktop and a popover slider for the mobile view. The revealer does sound good, but I will need some guidance on how to use GtkPopover properly because I'm always messing it up lol

jannuary:
Doesn't GTK have a [class](https://docs.gtk.org/gtk4/class.VolumeButton.html) specifically for that?

Diegovsky:
Thank you, it seems to be exactly what I was looking for. Also, which layout do you prefer?

xou816:
Hi! have you been able to look at the GtkVolumeButton? ping me if you need a review otherwise I'll hold off from it for now!

Diegovsky:
Will look into it. Good to know something like it exists :)

>  ping me if you need a review otherwise I'll hold off from it for now!

Sure

Amzd:
Any chance we can add a volume slider in the settings if we can't cut the knot on where it should be located on the player?

Diegovsky:
> Any chance we can add a volume slider in the settings if we can't cut the knot on where it should be located on the player?

Actually, would you mind to take it over from me? I don't have the time to work on it but I can review. In the end, the design is not what matters most, but the functionality.

Diegovsky:
Re-started work on this. Rebase was successful, however I need to migrate from those pesky .ui xml files to the more elegant blueprint files :)

Diegovsky:
By the way, [GtkVolumeButton is deprecated](https://docs.gtk.org/gtk4/class.VolumeButton.html). Not sure why.

Diegovsky:
image
What do you think?

I'm still unsure if this slider goes here or somewhere else, lol

Diegovsky:
This also depends on #684 to fix ci errors

UnlimitedCookies:
In my opinion there shouldn't be a popup like in celluloid, because the volume control is very essential to a music player and should be accessible quickly.  The best place to put it would be in my opinion to the right side of the player control area, like the proprietary spotify client currently does. This would probably suffice moving the time indicators.
image

Diegovsky:
I aggree, but I think it wouldn't look as good in a mobile view, though.

PR #531, opened 2022-06-21
Nest "More from artist" menu entries in submenu

supremesnickers:
Previously, every listed artist had its own "More from artist" menu entry, this commit nests them in a submenu with each entry representing an artist. Songs with a single artist are not affected.

Submenu:
screenshot1

Entries in submenu:
screenshot2

### DIFF
--- a/src/app/components/artist_details/artist_details_model.rs
+++ b/src/app/components/artist_details/artist_details_model.rs
@@ -123,11 +123,27 @@ impl PlaylistModel for ArtistDetailsModel {
 
         let menu = gio::Menu::new();
         menu.append(Some(&*labels::VIEW_ALBUM), Some("song.view_album"));
-        for artist in song.artists.iter().filter(|a| self.id != a.id) {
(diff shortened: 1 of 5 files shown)


src/app/components/artist_details/artist_details_model.rs
```diff
@@ -123,11 +123,27 @@ impl PlaylistModel for ArtistDetailsModel {
...
+            let submenu = gio::Menu::new();
+
+            for artist in artists_iter {
+                submenu.append(
+                    Some(&artist.name),
+                    Some(&format!("song.view_artist_{}", artist.id)),
+                );
+            }
+
+            menu.append_submenu(Some("More from"), &submenu);
```
xou816, line 139: these strings wont be translated -- you have to add a gettext call here, or I guess we could cheat and have `labels::more_from_label("")` here :see_no_evil:

xou816 (commented):
looking good :) sorry for the late review!

PR #546, opened 2022-09-11
po: Added Estonian translation

tenseoverflow:
Added Estonian (et) translation to the project.


xou816:
hi! sorry for the late response

have you added the translations to poeditor? i prefer updating all translations at once by pulling them from poeditor, its easier to manage possible conflicts that way :)

tenseoverflow:
alrighty! I'll add them there soon.

Issue #274, opened 2021-08-19
Provide an alternative to GNOME Keyring for saving token


Exagone313:
Currently, if GNOME Keyring is not installed, spot asks for account credentials each time it is started. Saving username and password in a configuration file should be avoided, that is why requiring GNOME Keyring is easier to develop as a secure by default pattern.

In #255, saving a token has been added to spot. This leads to the possibility of saving a less critical piece of credentials in a plain text file, with minimum security issues:
* the official Spotify client does that, and
* every web browser does that for cookies.

My suggestion is to provide a way to save this token in a plain text file (e.g. in `${XDG_CONFIG_HOME}` / `~/.config`).

Depending to how this feature may not be encouraged, this could be a feature disabled at compile time.

Currently, this prevents me from using this application, as I do not want to have GNOME Keyring hooking up every other app that supports it.

xou816:
Hi!
I am not sure storing that token would be enough, sadly, as it really short lived (an hour maybe?). Of course that'd be better than nothing, but still not great :/
Looks like we'll need to implement a proper, "official" auth flow one of these days, that does not rely on username+password! That we would have a refresh token which is better already.
Not trying to start a debate on GNOME Keyring, just curious, what do you mean by "hooking up every other app"?

xou816:
Possible solution to at least avoid plain text file: asymmetric encryption with perhaps a compile time random key? All of this behind an opt-in crate feature. This is not much safer, but that at least prevents accidentally accessing/indexing a plain text password file for instance.  

I am not a security expert though which is why I'd rather rely on GNOME Keyring and similar applications...

Diegovsky:
Just chiming in regarding where to store: `XDG_STATE_HOME` was recently added to the XDG spec to store this kind of thing.

Also, I don't think it should store the token since it's so short lived. Maybe spotify gives a refresh token? Asymmetric encryption is a good way to store it, +1 from me. However, randomly generating it at compile time will have the side effect of invalidating the current key when updating.

Diegovsky:
Oh yeah. Sometimes the app will crash and lose your info :(

xou816:
@sSoulllesSs have you tried the steps listed in the README? If it still does not work, please open a dedicated issue with more details for investigation (distro, logs, etc), this issue is not about the keyring feature but about providing a possible alternative to it. Thanks! :)

(also not aware of that bug you mention @Diegovsky  :/ a crash that causes keyring's credentials to be wiped?)

[...]

PR #560, opened 2022-10-31
Try to unlock the keyring collection when it is locked

rainDiX:
On KDE plasma, the secret service is initially locked even though the kwallet is unlocked. Currently if spot is the first application to require the wallet it will fail to access it. My current workaround is to open another app that unlock the wallet and then opening spot.
With this PR, spot will try to unlock the keyring collection if it is locked, resulting on a prompt asking the user for the keyring password, if the keyring unlocking fail, the behavior should be the same as before.
It should help #274 at least on the KDE side. I tested this PR on KDE, can you check if no regression are present on GNOME ?

### DIFF
--- a/src/app/credentials.rs
+++ b/src/app/credentials.rs
@@ -30,6 +30,9 @@ impl Credentials {
     pub fn retrieve() -> Result<Self, Error> {
         let service = SecretService::new(EncryptionType::Dh)?;
         let collection = service.get_default_collection()?;
+        if collection.is_locked()? {
+            collection.unlock()?;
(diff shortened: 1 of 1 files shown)


xou816 (approved):
lgtm, i'll see how it goes on gnome.

thanks for your contribution and for your patience!

xou816:
works perfectly, gnome shell shows a prompt if the keyring is locked. its a bit surprising perhaps, but better than what we have now :) i assume it should be possible to customize the prompt, i'll look into it someday

PR #590, opened 2023-02-12
Playlist edition

xou816:
i honestly dont remember what was left to do here


Diegovsky:
I'll pull it to look into :)

Diegovsky:
Don't know the scope of the change, but a remove track button is missing

xou816:
so the way I see it is the three-dots menu is for non-actions, just obtaining more info about a track. I'm thinking of removing the queue option that's been here for a while eventually

instead I think actions should be implemented with the selection mode that we have, so that things can be done in bulk!

removing a track from a playlist works by entering selection mode, selecting one or multiple songs and using the appropriate option at the bottom; similarly adding one or more tracks is done with the selection mode

also I wish there was some way to make this discoverable, but long pressing a track enables selection mode + selects that track, so it's not all that much slower than using the three-dots menu

Diegovsky:
I totally aggree, it is more efficient to do stuff in bulk, though, I like the option for the quick one-offs to be as "fast" as a complex operation. Totally up to you though, I'm fine with the current implementation :)

src/app/components/sidebar/sidebar_item.rs
```diff
@@ -0,0 +1,168 @@
...
+            Self::SavedTracks => "starred-symbolic",
+            Self::NowPlaying => "music-queue-symbolic",
+            Self::SavedPlaylists => "view-app-grid-symbolic",
+            Self::Playlist(_) => "playlist2-symbolic",
+        }
+    }
+}
+
+impl SidebarItem {
+    pub fn for_destination(dest: SidebarDestination) -> Self {
```
Diegovsky, line 62: I would name this `from_destionation` to be more consistent with rust naming conventions

src/main.rs
```diff
@@ -146,8 +146,18 @@ fn register_actions(app: &gtk::Application, sender: UnboundedSender<AppAction>)
     app.add_action(&make_action(
         "search",
         AppAction::BrowserAction(BrowserAction::NavigationPush(ScreenName::Search)),
-        sender,
+        sender.clone(),
     ));
+    let action = SimpleAction::new("open_playlist", Option::from(glib::VariantTy::STRING));
+    action.set_enabled(true);
+    action.connect_activate(move |_, playlist_id| {
+        // TODO: clean up unwraps
```
Diegovsky, line 154: Should probably fix this TODO before merging

xou816:
thanks for the review I appreciate it!!!

almost done here I think, save for a few details, but I want to merge it soon enough as it does rework a few important things

Diegovsky:
No problem! I hope I'm doing it right :)

xou816:
yeah I definitely appreciate the feedback! :)

Issue #538, opened 2022-07-11
Overlap between timestamp and long song title


s-brekke:
**Describe the bug**
In long song names, the title of the song overlaps the length of the song written to the right. One solution could be to present the length of the song in the same size and vertical position as the name of the artist.

**To Reproduce**
1. Play a song with a long name, or make the window narrow enough.

**Expected behavior**
Text should not overlap.

**Screenshots**
![bilde](https://user-images.githubusercontent.com/67145841/178216643-65e18ec3-60b8-4b34-b05e-7c13ed3d1a13.png)

**General information:**
 - Fedora Workstation 36
 - Flathub
 - Version 0.3.3

PR #597, opened 2023-02-16
fix overlap between timestamp and long song title

XGZepto:
Fix [issue#538](https://github.com/xou816/spot/issues/538).

`column-span` for song_title was set to `3`, led to the overlap with song_length. It is fixed by reducing the field to `2`.

## Screenshot

image

### DIFF
--- a/src/app/components/playlist/song.ui
+++ b/src/app/components/playlist/song.ui
@@ -67,7 +67,7 @@
         <property name="yalign">1</property>
         <property name="hexpand">1</property>
         <layout>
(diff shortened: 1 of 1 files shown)


Diegovsky:
Simple one line change, LGTM. Let's see what @xou816 thinks about this UI change.

xou816:
sure yeah if you're confident with small changes feel free to review and even merge!

Diegovsky:
Ok, then, I'll press the button !

Diegovsky (approved):

PR #599, opened 2023-02-19
Resolve #539

### LINKED ISSUE #539: Non-Standard Behavior in Rewind Button (intentional?)
**Describe the bug**
When invoking the "rewind"/previous track command (either through the Spot UI or dedicated keyboard buttons), Spot will skip directly to the previous song in the playlist. AFAIK this is non-standard behavior for a music player; typically the rewind command skips to the start of the current song unless it's only a few seconds in, in which case *then* it goes to the previous song.

**General information:**
 - Distribution: Fedora 36
 - Installation method [e. g. built from source, installed from Flathub...]: Flathub
 - Version [e.g. 0.1.0]: 0.3.3
 - Device used [e. g. desktop, phone...]: Desktop

I'd also like to note that I'm willing to try my hand at changing this myself, as a way to get more familiar with Rust. I just wanted to open an issue first to ensure this wasn't an intentional design choice.

SomewhereOutInSpace:
This PR corrects the behavior of the rewind/"previous song" action to match convention and official Spotify clients. 

With this patch applied, rewinding will only jump to the previous track if the current track has been playing for no more than 2 seconds. Otherwise, playback simply seeks to the start of the current track.

Notes:
- I copy-pasted the `PositionMicros` definition and `impl`s from the MPRIS code; you could definitely hoist them out of their respective modules and avoid the duplication, but I figured it would be alright - rule of three and all that.
- My code seems to hold up under testing, even under vigorous scrubbing and other player abuse. However, I'm not 100% clear on the exact effects of a `Load` event or the difference between a `Seek` and a `SyncSeek`, so there might be a logic error lurking in there.

### DIFF
--- a/src/app/state/playback_state.rs
+++ b/src/app/state/playback_state.rs
@@ -158,17 +158,22 @@ impl PlaybackState {
     fn stop(&mut self) {
         self.list_position = None;
         self.is_playing = false;
+        self.seek_position.set(0, false);
     }
 
     fn play_index(&mut self, index: usize) -> Option<String> {
         self.is_playing = true;
         self.list_position.replace(index);
+        self.seek_position.set(0, true);
         self.index.next_until(index + 1);
         self.current_song_id()
     }
 
     fn play_next(&mut self) -> Option<String> {
-        self.next_index().and_then(move |i| self.play_index(i))
+        self.next_index().and_then(move |i| {
+            self.seek_position.set(0, true);
+            self.play_index(i)
+        })
     }
 
     pub fn next_index(&self) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


src/app/state/playback_state.rs
```diff
@@ -292,8 +295,10 @@ impl UpdatableState for PlaybackState {
             PlaybackAction::TogglePlay => {
                 if let Some(playing) = self.toggle_play() {
                     if playing {
+                        self.seek_position.resume();
```
xou816, line 298: (applies to other lines below)

let's try to keep as much of the state manipulation inside the various methods above (toggle_play in that particular case) so that the already quite large match does not get larger!

src/app/state/playback_state.rs
```diff
@@ -343,20 +351,31 @@ impl UpdatableState for PlaybackState {
...
+                self.seek_position.set(0, false);
                 vec![PlaybackEvent::PlaybackStopped]
             }
             PlaybackAction::Previous => {
-                if let Some(id) = self.play_prev() {
+                // Microseconds -> seconds conversion, divide by 1 million.
+                // Only jump to the previous track if we aren't more than 2 seconds into the current song.
+                // (This replicates the behavior of official Spotify clients.)
+                if self.prev_index().is_some() && (self.seek_position.current() / 1_000_000) <= 2 {
+                    let Some(id) = self.play_prev() else {
```
xou816, line 362: cool, hadn't played with the new let ... else myself yet :)

xou816 (commented):
pretty clean and straightforward! thanks for the contribution!!

as for PositionMicros: I would just merge both eventually. but if you want to keep two versions that's fine -- in that case maybe modifying it to work with milliseconds is interesting; MPRIS requires microseconds but we might not want to be that specific

src/app/state/playback_state.rs
SomewhereOutInSpace: Excellent point! I've pushed a commit that hoists everything apart from seeking into the appropriate methods. 

(I'm gonna blame me being a fish out of water with Builder for not thinking of this :p)

xou816 (approved):
LGTM! :)

xou816:
wait, I didnt properly re-test but something is up with the unit conversions :s

PR #610, opened 2023-02-24
Port UI to Blueprint

### LINKED ISSUE #606: Consider porting UI to Blueprint
Blueprint (https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/) greatly improves the readability of `.ui` by reducing the verbosity of the syntax, as well as having better integration with IDEs.

I've done a port of almost all `.ui` files here https://github.com/alissonlauffer/spot/compare/3e620b21344ad4908ed10239ce9bd0d70dee3e96...976fd2d79853bbc12e32b099424c9ea9e4e6b400 if you want to see the diff, but I didn't open a PR because I wanted to know your opinion, and we need to coordinate this task since it would make all PRs that touch `.ui` files conflict.

alissonlauffer:
This PR does the following:
 - Port all `.ui` files to Blueprint.
 - Add Blueprint support for Flatpak.
 - Enable gtk-rs Blueprint validation feature.
 - Manually port files that the porter couldn't port automatically.
 - Update .pot and .po files.

Closes #606


alissonlauffer:
The reason for this PR being marked as "draft" is that there are still two unresolved things:
 - `src/app/components/login/login.ui` couldn't be ported because it uses GTK attributes, which are not supported by Blueprint, and [Blueprint docs](https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/examples.html#gtk-label) says the following:
   - _"Blueprint does not have syntax for attributes on Gtk.Label. Instead, you should use [platform classes](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/style-classes.html), [CSS](https://docs.gtk.org/gtk4/css-properties.html), [Pango Markup](https://docs.gtk.org/Pango/pango_markup.html) or [Gtk.Label.set_attributes](https://docs.gtk.org/gtk4/method.Label.set_attributes.html)."_
 - The `GtkAboutDialog` in `src/window.ui.in` should probably be migrated to an `AdwAboutWindow` inside the code itself(?) instead of using templates, as they are not supported in Blueprint too.

xou816:
looking good!! 

>  `src/app/components/login/login.ui` couldn't be ported because it uses GTK attributes, which are not supported by Blueprint, 

I see! The attributes on the title don't seem all that useful, we can ignore them; as for the error message, we can use pango markup (`<b></b>`) or add a css class.

>  The `GtkAboutDialog` in `src/window.ui.in` should probably be migrated to an `AdwAboutWindow` inside the code itself(?) instead of using @@ placeholders, as they are not supported in Blueprint too.

There is no special support needed, the `@VAR@` things are preprocessed by meson:

```
window_ui = configure_file(
    input: 'window.ui.in',
    output: 'window.ui',
    configuration: conf
)
```

porting to `AdwAboutWindow` sounds like a good idea -- for another issue though :)

(Also, the POTFILES file should be updated)

alissonlauffer:
I have now ported `login.ui` to Blueprint, using CSS instead of attributes.

About the `window.ui.in` file, it seems that what's being done with XML cannot be done with Blueprint, since it does not accept the format used in XML UI files for items (newlines).

xou816:
> About the `window.ui.in` file, it seems that what is being done with XML cannot be done with Blueprint, since it does not accept the format used in XML UI files for items (newlines).
> 
> Personally, I would just keep this file not ported for now, until someone upgrades that `GtkAboutDialog` with an `AdwAboutWindow`.

Switching to `AdwAboutWindow` won't automatically fix it, at the moment we still inject the list of contributors, the version, etc, at build time with meson. By the looks of it, blueprints supports newlines but we might have to alter a little bit the meson configuration :) 

image

If you're willing to go through with this I can help of course! I'm hopeful we won't have much trouble with that.

> EDIT: About translations, what gettext command should I run to update the translations? It's been a long time since I last used gettext.

First the POTFILES file should reference all blueprints and then you can run
`ninja spot-pot -C target`

If all goes well the changes to the .pot file should be minimal :) (just things moving around a bit at worst)

alissonlauffer:
Hello, I apologize for not seeing the commands for updating the .po files, sometimes I think I'm blind :)

As for the about window, I did try to make it work with Blueprint, but it failed to compile. I was able to show a window with the translator-credits parameter, but not with authors (it only shows if I remove the authors entry):
image

Otherwise, this PR should be good to go.

xou816:
No worries at all :)

Hmm ok interesting I'll have a look at this dialog issue!

xou816:
Looks like it's a limitation in blueprint for now https://gitlab.gnome.org/jwestman/blueprint-compiler/-/issues/38

What we can do is set those properties from rust code instead and start by leaving them empty in the blueprint, it should compile then. Let me think about it a little, I would really like to get rid of that big window.ui file if we're going the blueprint route :)

xou816:
if you don't mind rebasing a little your work, I'll merge #617 which cleans up the meson.build files and moves the inclusions of authors/artists/translator credits on the rust side, so it should be easier to port the last file

po/POTFILES
```diff
@@ -15,31 +15,31 @@ src/app/components/user_menu/user_menu.rs
 src/app/state/login_state.rs
 src/main.rs
 
-# find src -name "*.ui*" -print
+# find src -name "*.ui*" -o -name "*.blp*"
 src/window.ui
```
xou816, line 19: .blp now right? :tada: 

(also would you mind leaving the po files in their previous state? I regenerate them from poeditor every one in a while, I'll probably do it soon anyway to make sure nothing broke as it is a bit hard to review them here)

.gitignore
```diff
@@ -3,4 +3,5 @@
 /cargo/
 /build/
 /.flatpak-builder/
-/src/config.rs
\ No newline at end of file
+/src/config.rs
+/subprojects/blueprint-compiler
```
xou816, line 7: did you end up adding the subproject in the end? I don't see it in the tree (the wrap file)

(not that we absolutely need to)

po/POTFILES
alissonlauffer: Yes, but I have kept *.ui* in the find command in case we need to use .ui files again.

alissonlauffer: I'll rebase my PR and remove the edits I made to .pot and .po files once I get home.

xou816: sorry I meant the `src/window.ui` line (19)

(the find command was there just to help me remember how to quickly list all ui (now blp) files, we might not need to have a command that helps with porting here, but it's fine)

xou816:
looking good to me, lets merge :) thanks a lot!

xou816 (approved):

PR #612, opened 2023-02-25
Fix autologin when password contains backslashes

### LINKED ISSUE #596: Autologin doesn't work
**Describe the bug**
Autologin doesn't work. Every time I open Spot it prompts me to log back in. If I close the login prompt, it refuses to play saying that the player isn't responding. GNOME Keyring is unlocked and the "Spotify Credentials" entry is present.

**To Reproduce**
1. Open Spot and log in
2. Close Spot
3. Open Spot again
4. See the login prompt
5. If you close the login prompt without logging in, you can't play any tracks

**Expected behavior**
Autologin should work.

**General information:**
 - Distribution: Artix Linux (+ Sway Wayland compositor)
 - Installation method: Built from source via AUR package `spot-client-git`
 - Version: Commit 4b1209dc97daa5ba57be0178462baf388d014fc5 from `development` branch
 - Device used: Desktop

**Stack trace:**

There's no stack trace, but this is the output when running `spot` from the terminal (although it doesn't look relevant):

```
(process:8927): Adwaita-CRITICAL **: 21:43:21.017: adw_leaflet_set_visible_child: assertion 'contains_child' failed
```

**Additional context**
I'm not using Flatpak.

On a Fedora installation, it works fine with the latest Flatpak version.

samsapti:
This PR fixes #596. I also changed the error type returned by `Credentials::retrieve()` from `Error::Unavailable` to `Error::NoResult`, as it seems more appropriate for when the secret can't be read.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@ Warren Hu
 bbb651
 Julius Rüberg
 janbrummer
+Sam Al-Sapti
--- a/src/app/credentials.rs
+++ b/src/app/credentials.rs
@@ -35,7 +35,7 @@ impl Credentials {
         }
         let items = collection.search_items(make_attributes()).await?;
         let item = items.get(0).ok_or(Error::NoResult)?.get_secret().await?;
-        // We need to escape backslashes in order to correctnly deserialize passwords that contain backslashes
+        // We need to escape backslashes in order to correctly deserialize passwords that contain backslashes
         let secret = String::from_utf8(item)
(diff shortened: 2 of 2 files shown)


samsapti:
We might wanna refactor this, as it doesn't seem like the right solution. I think the issue lies with how `serde_json` serializes the `Credentials` struct.

samsapti:
Closing as the issue is unrelated to Spot.

PR #642, opened 2023-03-31
Fix login screen according to #625

### LINKED ISSUE #625: Minor label changes for the new login screen
*Note: This is a low-priority issue, but excellent for newcomers!*

The new login screen is coming together nicely! :tada:

new-login

I have a few things on the label/string side that I think could be changed:

- The "Login with your Spotify Account" label beneath the title could be justified to the center, to match the rest of the text on the page. I think this just requires adding `justify: center;` to the label in the login blueprint file, but my building skills aren't quite there yet, so I'm not able to test it.

- "Login" should be "Log in". I had to search that up. "Login" is a noun/adjective ("the login window"), "log in" is a verb ("I tried to log in"). Also, "to use Spot" could be changed to "to use the app", so we avoid the slightly awkward repetition of the app name from the "Welcome to Spot" label. Or we could just drop the "to use x" part altogether, as it's kinda implied from the fact that you're trying to log in. This is my suggestion for the label:
    > Log in with your Spotify account. A Spotify Premium subscription is required to use the app.

- I don’t have an easy-to-remember Spotify username, as it was automatically generated when we created our family subscription, but I’m able to log in with my email in the username field. We should specify that that’s possible as well. We can do this by changing the "Username" label to "Username or Email". Another web search reveals that both "E-mail" and "Email" are acceptable writing methods, but Email is the most common one. The most important thing is that it's Header Capitalized in accordance with the HIG :grin: 

I'm going to stop annoying you with my design nitpicking for now, cheers :sweat_smile:

hegdenischay:
I have been a user for a couple years now, so I figured contributing wouldn't hurt.

The commit is pretty much verbatim to the description of the issue.

Here's the new login screen that this PR creates: 

new login screen

### DIFF
--- a/po/spot.pot
+++ b/po/spot.pot
@@ -332,13 +332,13 @@ msgstr ""
 #. Translators: Login window title, must mention Premium (a premium account is required).
 #: src/app/components/login/login.blp:53
 msgid ""
-"Login with your Spotify Account. A Spotify Premium subscription is required "
-"to use Spot."
+"Log in with your Spotify Account. A Spotify Premium subscription is required "
+"to use the app."
 msgstr ""
 
 #. Translators: Placeholder for the username field
 #: src/app/components/login/login.blp:66
-msgid "Username"
+msgid "Username or Email"
 msgstr ""
 
 #. Translators: Placeholder for the password field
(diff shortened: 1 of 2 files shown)


xou816 (approved):
looks good to me, thanks for contributing!! :)

PR #643, opened 2023-03-31
Use libadwaita::AboutWindow instead of gtk::AboutDialog

### LINKED ISSUE #636: Use AdwAboutWindow
Adwaita has an [About Window](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AboutWindow.html) widget that Spot should adopt. It looks better and is better organized than GTK's about window. Should be pretty easy to switch to as well.

hegdenischay:
Switched to libadwaita's AboutWindow as mentioned in Issue #636.

I believe I've done it correctly, but I'm still quite the n00b in GTK4 and Rust, so here's a screenshot of the new about page that this PR creates: 

new about screen

### DIFF
--- a/src/app/components/user_menu/user_menu.rs
+++ b/src/app/components/user_menu/user_menu.rs
@@ -16,7 +16,7 @@ impl UserMenu {
     pub fn new(
         user_button: gtk::MenuButton,
         settings: Settings,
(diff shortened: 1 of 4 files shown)


xou816 (approved):
lgtm :) thanks!

Issue #508, opened 2022-03-30
Library does not show followed artists


NicolasLM:
On the official Spotify application the way to save an artist to the library is to "follow" the artist. So I tend to follow artists rather than add individual albums or tracks to my library.

This means that on Spot my library is completely blank, I guess similar to a brand new account.

dallenbaldwin:
I was also curious about this and did some quick looking. It seems like there is an endpoint in the [Web API](https://developer.spotify.com/documentation/web-api/reference/#/operations/get-followed) reference

PR #671, opened 2023-07-13
WIP: Show followed artists in the sidebar

patgruber:
These changes show the artists that the user follows in the sidebar under the saved playlists.
The used API call is https://developer.spotify.com/documentation/web-api/reference/get-followed and the issue related to this is https://github.com/xou816/spot/issues/508
The issue describes showing followed artists in the library view, but here they are shown in the sidebar similar to how the Spotify app does it.


patgruber:
One thing that still needs work is the pagination of the request, since it is slightly different than using limit and offset.
Also the caching client doesn't cache the result yet

patgruber:
Sadly I won't be able to work on this anymore, since I don't have a Spotify Premium account anymore. 
Anyone is welcome to pick it up and finish it.

Issue #591, opened 2023-02-12
⚠️ Current maintenance policy -- read for issues and PR submissions


xou816:
(work in progress)

### Submitting issues/feature requests

There's a bit (ahem) of catching up to do regarding issues and pull requests. Please allow us to work through it! **Bug fixes will be prioritized over new features for the time being.**

**For reviews, please involve one of the following people** and allow them time to review, as we're all busy people :)
> @xou816 @Diegovsky 

For new features that you'd like to implement yourself, **please start with an issue** so we can flesh out together which way to go. This will save time for everyone :)

___________________________________________

### Maintainers

If you'd like to start/stop being involved in some of the maintenance effort, please signal yourself below, I'll add/remove you to the list :) 

Tasks we could use some help with:
- reviewing PRs;
- updating translations (pulling them from POEditor; to be better documented);
- fixing dependabot PRs when that's needed;
- fixing issues;
- triaging issues and labeling them, trying to reproduce them;
- preparing releases (to be documented);
- contributing new code in general, of course, but I´d like this to be lower priority while we catch up on the rest.

I'm still thinking about how to handle permissions. For now, it's likely that I'll still have to be the one to hit merge -- but I'll try to defer as much as possible to maintainers and trust your reviews.

xou816:
this week I'll try to go through open PRs that seem close to ready and then bugs

there are a couple dependencies update to handle but nothing too bad!

we need to have a look at pull transations from poeditor as well :)

samsapti:
@xou816 just a small suggestion: Maybe you could give co-maintainers write access to the repo, but only you have write access to the `development` and `master` branches? That would be easy to do with branch protection rules.

Diegovsky:
>  For now, it's likely that I'll still have to be the one to hit merge

I'm all abord with this! IMHO your vision of the project is what goes :)

knokelmaat:
Hey, I am the one who put op the post on /r/gnome. I would also like to contribute if needed. I don't have much experience with Github, but I'll see what I can do. I'll start by just cloning and getting everything working on my end :smile: .

Also, might it be interesting to have some sort of chat room for quick development conversations (like asking for help with a certain error or stuff like that)? Going through github issues might be a bit of a barrier for small things that are just needed to get going.

I would certainly like to help accommodating that! I quickly made a space on matrix for example:
https://matrix.to/#/#spot-devel:matrix.org

samsapti:
[...]

PR #677, opened 2023-08-24
Improve cover button display in library

panurg:
If there are few albums in library then cover button expands down to the box bottom border. That does not happen in other places for some reason. Also scrollbar in library should not be always displayed, as it does not make sense for the case with few albums.
Before:
Screenshot from 2023-08-01 01-36-52
After:
Screenshot from 2023-08-01 02-24-03

### DIFF
--- a/src/app/components/album/album.blp
+++ b/src/app/components/album/album.blp
@@ -5,6 +5,7 @@ template $AlbumWidget : Adw.Bin {
   Button cover_btn {
     hexpand: false;
     halign: center;
+    valign: start;
 
     Box {
       halign: center;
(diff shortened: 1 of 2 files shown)


panurg:
ci-check fails in `src/connect/player.rs` Just FYI...

Diegovsky:
Hey there! Thank you for the fix. I'll be looking into this

Diegovsky:
wow i should have looked into this sooner, it's a simple fix

Diegovsky:
Pushed this to my fork of the project. See https://github.com/xou816/spot/issues/591#issuecomment-1895880929 for more information.

Thank you for your contribution :)

PR #684, opened 2023-11-02
Fix ci error

Diegovsky:


Diegovsky:
Finally, it passes! @knokelmaat @xou816 could any of you please review the changes?

To summarize what I did: I updated ALL `Cargo.lock` dependencies with `cargo update`. I don't see a reason not to since we have a bunch of tests and semver guarantees it is fine to do so.

Also, updating in general is good for security. Some of the security problems raised by dependabot were solved just by doing so, I think.

nmschulte (approved):

Diegovsky:
Thank you for the review @nmschulte. Would you happen to know why it can't be merged yet despite you approving it? (I'm not very experienced when it comes to PRs...)

nmschulte:
> Would you happen to know why it can't be merged yet despite you approving it? (I'm not very experienced when it comes to PRs...)

Because I'm just a rando wizard with no power here, unfortunately.  We need some greybeards ... https://www.youtube.com/watch?v=m_mPE9gcQJo

Diegovsky:
That's unfortunate

knokelmaat (approved):

oknyshuk:
it still doesn't pass for me on Arch, rust 1.74.0. here's the testlog: https://0x0.st/HvRW.txt

Issue #738, opened 2024-12-08
No playlists show on sidebar


davidosdr:
**Describe the bug**

After successfully login from the browser into the app, no playlists appear on the sidebar. A popup says check the logs.

**To Reproduce**

Install flathub version 0.5.0

Login from browser

Enter the app

**Expected behavior**

Playlist are supposed to be on the sidebar

**Screenshots**

**General information:**
 - Distribution: ZorinOS 17 (Ubuntu)
 - Installation method [e. g. built from source, installed from Flathub...]: Flathub
 - Version [e.g. 0.1.0]:  0.5.0
 - Device used [e. g. desktop, phone...]: Desktop

**Stack trace:**
If applicable, run the application from a terminal and paste relevant log output.
```
flatpak run --env=RUST_BACKTRACE=full dev.alextren.Spot


(process:2): Adwaita-CRITICAL **: 20:57:42.889: adw_leaflet_set_visible_child: assertion 'contains_child' failed
MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:763: FINISHME: support YUV colorspace with DRM format modifiers
MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:794: FINISHME: support more multi-planar formats with DRM modifiers
[src/api/cached_client.rs:259:17] &cache_key = "me_playlists_0_30.json"
[src/api/cached_client.rs:259:17] e = Error("invalid type: null, expected a sequence", line: 1, column: 20590)
[2024-12-08T19:57:45Z ERROR spot::app::components] Spotify API error: invalid type: null, expected a sequence at line 1 column 20590
[src/api/cached_client.rs:259:17] &cache_key = "me_playlists_0_30.json"
[src/api/cached_client.rs:259:17] e = Error("invalid type: null, expected a sequence", line: 1, column: 20590)
[2024-12-08T19:57:47Z ERROR spot::app::components] Spotify API error: invalid type: null, expected a sequence at line 1 column 20590
[src/api/cached_client.rs:259:17] &cache_key = "me_playlists_0_30.json"
[src/api/cached_client.rs:259:17] e = Error("invalid type: null, expected a sequence", line: 1, column: 20590)
[2024-12-08T19:57:49Z ERROR spot::app::components] Spotify API error: invalid type: null, expected a sequence at line 1 column 20590

```

**Additional context**

I use ZorinOS 17, ubuntu-based distro.

shrapnelnet:
[...]

PR #745, opened 2025-04-15
Fix: handle empty playlist images from API response

danielhauser:
Hi,
I tried to fix the bug in #738. The problem is that the Web API may return null for images (That happens if the playlist doesn't have any songs yet for example). The behaviour is described in their [API reference](https://developer.spotify.com/documentation/web-api/reference/get-playlist). I tested it locally and for me the changes do the trick. I used a default Icon as a place holder, should be fine I think. 

As I am not really experienced in Rust nor in contributing to public code I would very welcome constructive feedback :)

### DIFF
--- a/src/api/api_models.rs
+++ b/src/api/api_models.rs
@@ -173,7 +173,7 @@ trait WithImages {
             .map(|image| (criterion(image), image))
             .collect::<Vec<(T, &Image)>>();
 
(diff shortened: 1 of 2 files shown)


Diegovsky:
Can you test it your issue still happens in [maintenance](https://github.com/xou816/spot/tree/maintenance)?

PR #752, opened 2025-06-13
Add multi-arch Flatpak CI builds with ARM64 support

### LINKED ISSUE #751: The flatpak version is missing aarch64/stable ref, thus I can't instal 0.5.0 with oauth on postmarketOS
Only 0.4.1 version is availbale since it's the last version with that ref present but it no longer works and just throws "Invalid crredentials"

Tried compiling but gave up due to bunch of missing static libraries since postmarketOS is based on Alpine Linux

SolsticeSpectrum:
Fixes https://github.com/xou816/spot/issues/751

Testing is requried, it automatically creates releases using the softprops/action-gh-release github action
I tested on postmarketOS, altough I had to run these commands first to fix GPG issues and install gnome 47

```
flatpak remote-delete flathub
flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
```
```
flatpak install flathub org.gnome.Platform//47
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,72 @@
+name: spot-flatpak-multiarch
+
+on:
+  push:
+    branches: [master, development]
+    paths-ignore:
+      - '*.md'
+  pull_request:
+    branches: [master, development]
+  workflow_dispatch:
+
+jobs:
+  flatpak:
+    name: "Flatpak ${{ matrix.variant.arch }}"
+    container:
+      image: ghcr.io/flathub-infra/flatpak-github-actions:gnome-47
+      options: --privileged
+    strategy:
+      matrix:
+        variant:
+          - arch: x86_64
+            runner: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


Diegovsky:
I'm a bit on the fence here regarding merging this as I have no expertise with CI and don't have arm devices to test it out. Are you willing to become a contributor and help out with this? If so, I wouldn't mind.

SolsticeSpectrum:
> I'm a bit on the fence here regarding merging this as I have no expertise with CI and don't have arm devices to test it out. Are you willing to become a contributor and help out with this? If so, I wouldn't mind.

sure, I have no problem with that. I have been using the build for quite some time now and have no issues.

Diegovsky:
This is really cool! Development has shifted into https://github.com/Diegovsky/riff for now. I have disabled CI because I don't have experience with it and it was running every time despite being broken, but I am willing to accept arm support when we have CI up again!

